PR #1, opened 2021-09-30
Stage1 updated

morcuended:


morcuended:
This is way too outdated now, I will open a new PR at some point.

PR #34, opened 2021-11-18
Improvements on provenance tracking

Bultako:
The content of this PR is the following:

- fixes some issues derived from changes in the data processing steps that have not been translated into the model up to now.
- adds calibration processes to the simulate processing script
- tracks and saves python packages dependencies of the conda environment
- adds `lstchain_dl1ab ` process into the data sequence provenance
- adds `dl1_datacheck` process into the data sequence provenance
- adds missing DL2 merged file to the provenance
- adds lstosa config file used (i.e. `sequencer.cfg`) as a session parameter

The final graph produced is as below.
[provDLshort.pdf](https://github.com/gae-ucm/lstosa/files/7564674/provDLshort.pdf)


morcuended:
@bultako, I tried to resolve the conflicts. I have now a problem in simulate processing

```python
  File "/fefs/aswg/workspace/daniel.morcuende/software/lstosa/osa/scripts/simulate_processing.py", line 141, in simulate_subrun_processing
    run_str, subrun_idx = args[14].split(".")
ValueError: not enough values to unpack (expected 2, got 1)
```

We could add a unit test for `simulate_processing` and then for `provprocess` to check both scripts.

Bultako:
@morcuended 

Simulate processing should be fixed now.
I would add tests for `simulate_processing` and `provprocess` in another PR, as well as tests for `provenance/capture.py` and others improving code coverage.

morcuended:
Thanks, @Bultako I will have a look at the PR. Implementing the tests in another PR is fine with me.

morcuended:
Right now, with the entire RunSummary from 2020-01-17, simulate_processing takes >1 hour so we will need to shorten the list of runs to be simulated for the unit tests (I've already done it)

I tried `simulate_processing` + `provprocess` and found some things that could be improved:

- RunSummaryFile filepath right now indicates the basename of the file. Should be the absolute path.
- PointingFile filepath indicated a PosixPath. I think it should display just the string of that Path.
- MuonsCollectionRun filepath points to the `running_analysis/<date>` directory. However, it should point to `DL1/<date>` which is the final destination directory for the muons files. It should point to the same filepath as for the DL1Collection.

[r0_to_dl2_01808_prov.pdf](https://github.com/gae-ucm/lstosa/files/7614144/r0_to_dl2_01808_prov.pdf)

Other than that, it looks good to me.

osa/provenance/config/definition.yaml
```diff
@@ -56,21 +56,6 @@ activities:
         description:
             "Create DL1 files for an observation run and subrun"
         parameters:
-            - name: ObservationRun
```
morcuended, line 59: Are parameters no longer needed for some of the activities (eg `r0_to_dl1`, `dl1_to_dl2`)?

osa/provenance/utils.py
```diff
@@ -29,31 +29,39 @@ def parse_variables(class_instance):
...
+        # time_calibration  [2] .../20191124/v00/time_calibration.Run01625.0000.hdf5
+        # drivefile         [3] .../DrivePositioning/drive_log_20_02_18.txt
+        # runsummary_file   [4] .../RunSummary/RunSummary_20200101.ecsv
         # run_str           [5] 02006.0000
-        # historyfile       [6] /fefs/aswg/data/real/running_analysis/20200218/v0.4.3_v00/sequence_LST1_02006.0000.history
 
-        class_instance.AnalysisConfigFile = configfile
+        class_instance.ObservationRun = class_instance.args[5].split(".")[0]
+        class_instance.ObservationDate = nightdir
+        class_instance.SoftwareVersion = options.lstchain_version
```
morcuended, line 54: maybe we could use `lstchain.__version__` to avoid the usage of `options`

osa/provenance/utils.py
```diff
@@ -62,69 +70,92 @@ def parse_variables(class_instance):
             calibration_path / timecalibration_filename
         )
         class_instance.PointingFile = class_instance.args[3]
-        class_instance.ObservationRun = class_instance.args[5].split(".")[0]
-        class_instance.ObservationSubRun = class_instance.args[5].split(".")[0]
+        class_instance.RunSummaryFile = os.path.basename(class_instance.args[4])
```
morcuended, line 73: I think here we could just set `str(class_instance.args[4])` in order to show the abs path

osa/provenance/utils.py
```diff
@@ -62,69 +70,92 @@ def parse_variables(class_instance):
...
+        class_instance.ObservationDate = nightdir
+        class_instance.SoftwareVersion = options.lstchain_version
         class_instance.session_name = class_instance.ObservationRun
         class_instance.ProcessingConfigFile = options.configfile
 
+        class_instance.DL1SubrunDataset = (
+            f"{outdir_dl1}/dl1_LST-1.Run{class_instance.args[0]}.h5"
+        )
+        class_instance.MuonsSubrunDataset = (
+            f"{analysis_dir}/muons_LST-1.Run{class_instance.args[0]}.fits"
```
morcuended, line 108: {outdir_dir}

morcuended: Actually, these muons files are produced in the previous step `r0_to_dl1` so I guess `MuonsSubrunDataset` should be moved to the previous class instance.

osa/scripts/calibrationsequence.py
```diff
@@ -182,15 +182,15 @@ def drs4_pedestal(
     -------
     Return code
     """
+    if options.simulate:
+        return 0
+
```
morcuended, line 187: I actually moved the simulate option after building the commands for the scripts so I could test them. But I think it does not matter in the end.

osa/provenance/capture.py
```diff
@@ -605,6 +616,7 @@ def get_system_provenance():
             version=platform.python_version(),
             compiler=platform.python_compiler(),
             implementation=platform.python_implementation(),
+            packages=get_python_packages(),
```
morcuended, line 619: In which file can I see this python package info in the end?

morcuended (changes requested):
I would change some of the paths I indicated in the previous comment. Otherwise, it's fine with me.

osa/provenance/capture.py
Bultako: All system info is in the `processlevel_runid_prov.log` files, it is remove from from `json` and `pdf` graphs to improve readability.

osa/scripts/calibrationsequence.py
Bultako: I had to move it before the `get_input_file` call to avoid raised errors for non existing files.

osa/provenance/utils.py
Bultako: Ok.
Since these muons files are produced by `r0_to_dl1` I will add a `wasGeneratedBy` connection, keeping also the already existing `used` connection in `dl1_datacheck`.

osa/provenance/config/definition.yaml
Bultako: Yes, all parameters given are actually present in the paths of the produced datasets or as session attributes. (i.e. date, run number, software version..)

morcuended: I see, thanks

osa/scripts/calibrationsequence.py
morcuended: OK, fine with me as it is.

osa/provenance/utils.py
Bultako: I would actually replace `options.lstchain_version` everywhere in  `lstosa` by `get_lstchain_version()` as it is already done in `osa/utils/utils.py`

Bultako:
> I tried simulate_processing + provprocess and found some things that could be improved:

You don't need to run `rovprocess`if you use the `-p` flag.

osa/provenance/utils.py
morcuended: I agree. That function basically returns `f"v{lstchain.__version__}"`

I think that I implemented the options.lstchain_version to avoid importing lstchain several times, which takes several seconds. On the other hand, I want to get rid of the global `options` module at some point.

morcuended:
> > I tried simulate_processing + provprocess and found some things that could be improved:
> 
> You don't need to run `provprocess`if you use the `-p` flag.

Nice. I did not realize it before.

morcuended (approved):
@Bultako the PR looks good to me now. Shall we merge?

morcuended (approved):

PR #41, opened 2021-11-28
Shorten test run summary

morcuended:
In order to speed up simulate processing tests, is better to have a shorter run summary.


morcuended:
messed things up mixing branches :(

PR #44, opened 2021-11-29
Improve provenance code coverage

Bultako:
This PR improves provenance test code coverage.

`provenance/tests/test_capture.py` has been removed and provenance tests are now performed in `test_simulate_processing` function in `scripts/tests/test_osa_scripts.py` which cover most of the code present in the `provenance` folder, as well as lines in `simulate_processing.py` and `provprocess.py` scripts.


morcuended:
It seems that removing test_capture leaves provenance/utils.py untested (at least directly). Do you think it makes sense to partially test the actual content of the json provenance file?

Bultako:
> It seems that removing test_capture leaves provenance/utils.py untested (at least directly).

`provenance/utils.py` code is actually tested, with a code coverage of 92%

```
$ pytest --basetemp=test_osa osa/scripts/tests/test_osa_scripts.py -v -k test_simulate_processing --cov osa --cov-report html
$ open htmlcov/index.html
```

> Do you think it makes sense to partially test the actual content of the json provenance file?

That's a fair point..
I will include tests on file contents for `json` and `log` files produced in next commits in this PR.

morcuended:
> provenance/utils.py code is actually tested with test_simulate_processing producing a code coverage of 92%

You're right. Checking codecov webpage (whose bot by the way is no longer reporting in the PR as it used to do before) puzzled me.

Bultako:
> Checking codecov webpage (whose bot by the way is no longer reporting in the PR as it used to do before) puzzled me.

Yep, I think I should put `provenance/tests/test_capture.py` back to where it was so to have good codecov stats..

morcuended:
None of the things I tried worked. I had the impression that the problem was in the codecov.yml file but it seems nothing changes neither. I will merge this PR anyway and keep investigating the difference between pytest --cov-report and codecov % coverage.

Issue #54, opened 2021-12-13
Use onsite scripts from lstchain


morcuended:
We need to adapt lstosa calibration sequence to run these two files on a daily basis for a whole-dataset reprocessing:
- [onsite_create_calibration_file.py](https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/scripts/onsite/onsite_create_calibration_file.py)
- [onsite_create_drs4_pedestal_file.py](https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/scripts/onsite/onsite_create_drs4_pedestal_file.py)

Currently, we run directly the scripts/Tools from lstchain so it needs to be adapted.

Maybe it would be good to implement a separate script to not leave out the current `calibrationsequence.py`

morcuended:
@contrera @Bultako

@FrancaCassol confirmed this will be the scheme from now on:
- For massive reprocessings, lstosa will produce the calibration and drs4 pedestal files for each day. They will be based on the F-factor systematic file and DRS4 time calibration files already produced.
- For next-day processing calibration products will be produced already externally. lstosa will directly take the calibration files already produced.

PR #74, opened 2022-01-10
Add calibration to provenance tracking

Bultako:
This PR extends the provenance products files and graph produced to previous calibration steps (as shown in example prov graph [calibration_to_dl2_01808_prov.pdf](https://github.com/cta-observatory/lstosa/files/7838240/calibration_to_dl2_01808_prov.pdf)). It adds to the provenance tracking the calibration processes that create pedestal, time and charge calibration files used in already tracked process `ro_to_dl1`, as well as the check plots files produced. Consequently prov products files are now renamed:

- calibration_to_dl1_runID_prov.pdf
- calibration_to_dl2_runID_prov.pdf

For the moment, the calibration processes tracked are those involved in the **next day processing**, those using external lstchain onsite calibration scripts, only input and output files relative to those scripts are tracked skipping the rest of parameters hard coded in those scripts and outside the scope of lstosa. In the future, provenance tracking for **massive reprocessing** will be added. See https://github.com/cta-observatory/lstosa/issues/54#issuecomment-993253782 for more details.

It also improves and refactor existing provenance tracking code.

### DIFF
--- a/osa/provenance/config/definition.yaml
+++ b/osa/provenance/config/definition.yaml
@@ -74,6 +74,53 @@ activities:
               value: PedestalCheckPlot
 #              filepath: /fefs/aswg/data/real/calibration/20210913/v0.7.5/log/drs4_pedestal.Run02068.0000.pdf
 
+    calibrate_charge:
+        description:
+            "Create charge calibration file"
+        parameters:
+        usage:
+            - role: "Subrun for calibration"
+              description: "Raw observation file for calibration"
+              entityName: RawObservationFile
+              value: RawObservationFileCalibration
+#              filepath: /fefs/aswg/data/real/R0/20210913/LST-1.1.Run06274.0000.fits.fz
+            - role: "Pedestal file"
+              description: "Pedestal file used"
+              entityName: PedestalFile
+              value: PedestalFile
+#              filepath: /fefs/aswg/data/real/calibration/20210913/v0.7.5/drs4_pedestal.Run06268.0000.fits
+            - role: "Run summary"
+              description: "Run summary configuration"
+              entityName: RunSummaryFile
+              value: RunSummaryFile
+#              filepath: /fefs/aswg/data/real/monitoring/RunSummary/RunSummary_20210913.ecsv
+            - role: "Configuration file"
+              description: "Configuration file for camera"
+              entityName: AnalysisConfigFile
+              value: CalibrationConfigurationFile
+#              filepath: /fefs/aswg/software/virtual_env/ctasoft/cta-lstchain/lstchain/data/onsite_camera_calibration_param.json
+            - role: "Systematics correction file"
+              description: "Systematics correction file"
+              entityName: SystematicsCorrectionFile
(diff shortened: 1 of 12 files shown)


morcuended:
Thanks, @Bultako. I'll have a look.

osa/scripts/calibration_pipeline.py
```diff
@@ -90,14 +91,19 @@ def calibration_sequence(drs4_pedestal_run_id: str, pedcal_run_id: str) -> int:
...
+    drs4_pedestal_run_id: str, pedcal_run_id: str, history_file: Path
+) -> int:
     """
     Create a DRS4 pedestal file for baseline correction.
 
     Parameters
     ----------
     drs4_pedestal_run_id : str
         String with run number of the pedestal run
+    pedcal_run_id : str
```
morcuended, line 105: why adding this argument here? Is it needed for provenance?

Bultako: Yes.
It is needed to identify link each both two calibration tasks to a single calibration workflow identified with a label of two numbers, namely `drs4_pedestal_run_id-pedcal_run_id`.

morcuended:
Looks good to me. I'm going to test it in the container.

Bultako:
I think we have to define a boundary on which provenance info `lstosa` should keep. I'm tempted to say that all info defined outside `lstosa` should not be kept, so external scripts used like those from `lstchain` would appear as blackboxes with inputs and outputs files that may be also connected to other tasks in `lstosa`. The reason why to do this is to simplify maintenance. We could also decide to add the detailed provenance info happening in these external boxes, in that case there is now a lot of info missing, and any modif in those `lstchain` scripts used will need to modify the code for provenance tracking in `lstosa`.

morcuended:
> I think we have to define a boundary on which provenance info `lstosa` should keep. I'm tempted to say that all info defined outside `lstosa` should not be kept, so external scripts used like those from `lstchain` would appear as blackboxes with inputs and outputs files that may be also connected to other tasks in `lstosa`. The reason why to do this is to simplify maintenance. We could also decide to add the detailed provenance info happening in these external boxes, in that case there is now a lot of info missing, and any modif in those `lstchain` scripts used will need to modify the code for provenance tracking in `lstosa`.

Agreed. At least for now, we should at least keep track of which calibration files are used in r0_to_dl1 (which we have been doing so far). The information on the production of calibration files produced by osa is limited to the drs4 pedestal and calibration files and corresponding run numbers. We have no straight access to the time file (although in lstosa we look for this file in the same way for using it in r0 to dl1) nor the factor systematic file.

osa/provenance/config/definition.yaml
```diff
@@ -296,14 +368,26 @@ entities:
         description: "Pedestal file in FITS format on the disk"
         type: File
         contentType: application/fits
```
morcuended, line 370: Pedestal file is now produced in h5 format.

morcuended (approved):

PR #81, opened 2022-01-17
DL2 to DL3 step implementation

morcuended:
**Usage:** `dl3_stage -d 2021_08_08 -c cfg/sequencer.cfg LST1`

It makes use of the metadata information extracted from the TCU database (source name and RADec coordinates). (WARNING: a run catalog is to be implemented. This would ease the access to metadata information needed for the DL3 tool).

It pipes the lstchain scripts:
- `lstchain_create_irf_files` (once per selection cuts)
- `lstchain_create_dl3_file` (run-wise)
- `lstchain_create_dl3_index_files` (source-wise)

The idea is to end up with the following structure:

```
/fefs/aswg/data/real
├── monitoring
│   ├── RunSummary
│   ├── DrivePositioning
│   └── PixelCalibration
├── DL1
│   └── YYYYMMDD
│       └── vX.Y.Z
│           ├── muons.fits
│           └── tailcut84
│               ├── dl1.h5
│               └── datacheck_dl1.h5
├── DL2
│   └── YYYYMMDD
│       └── vX.Y.Z
│           └── tailcut84
│               └── dl2.h5
└── DL3
    └── YYYYMMDD
        └── vX.Y.Z
            └── tailcut84
                ├── std_cuts
                │   ├── irf_std_cuts.fits
                │   ├── source_name1
                │   │   ├── dl3_LST-1_Run00001.fits
                │   │   ├── dl3_LST-1_Run00002.fits
                │   │   ├── hdu-index.fits.gz
                │   │   └── obs-index.fits.gz
                │   └── source_name2
                │       ├── dl3_LST-1_Run00003.fits
                │       ├── dl3_LST-1_Run00004.fits
                │       ├── hdu-index.fits.gz
                │       └── obs-index.fits.gz
                └── other_cuts

```

Right now this script is intended to be run separately, once closer has been launched and files have been moved to final destinations.

TODO in further PRs: 
- Refactoring.
- DL3 stage should run right after the merging of DL2 files without having to depend on the closer. Sequencer should take care of this analysis step as well through the `datasequence`.
- Implement unit tests.
- Next-day high-level analysis.

### DIFF
--- a/cfg/sequencer.cfg
+++ b/cfg/sequencer.cfg
@@ -57,6 +57,12 @@ dl1ab_config: /software/lstchain/data/lstchain_standard_config.json
 # the MC DL2 uses a different config file, it can be set here.
 dl2_config: /software/lstchain/data/lstchain_standard_config.json
 rf_models: /data/models/prod5/zenith_20deg/20201023_v0.6.3
+dl3_config: /software/lstchain/data/dl3_std_config.json
+
+[IRF]
+mc_gamma: /fefs/aswg/data/mc/DL2/20200629_prod5_trans_80/gamma/zenith_20deg/south_pointing/20210923_v0.7.5_prod5_trans_80_dynamic_cleaning/off0.4deg/dl2_gamma_20deg_180deg_off0.4deg_20210923_v0.7.5_prod5_trans_80_dynamic_cleaning_testing.
+mc_proton: /fefs/aswg/data/mc/DL2/20200629_prod5_trans_80/proton/zenith_20deg/south_pointing/20210923_v0.7.5_prod5_trans_80_dynamic_cleaning/dl2_proton_20deg_180deg_20210923_v0.7.5_prod5_trans_80_dynamic_cleaning_testing.h5
+mc_electron: /fefs/aswg/data/mc/DL2/20200629_prod5_trans_80/electron/zenith_20deg/south_pointing/20210923_v0.7.5_prod5_trans_80_dynamic_cleaning/dl2_electron_20deg_180deg_20210923_v0.7.5_prod5_trans_80_dynamic_cleaning_testing.h5
 
 [SLURM]
 PARTITION_PEDCALIB: short
--- a/environment.yml
+++ b/environment.yml
@@ -21,6 +21,7 @@ dependencies:
   - h5py
   - joblib
   - traitlets=5.0
+  - click
   # dev dependencies
   - pre-commit
   - pytest
--- a/osa/configs/datamodel.py
+++ b/osa/configs/datamodel.py
@@ -18,7 +18,9 @@ class Telescope(Night):
 class Source(Telescope):
     def __init__(self):
         super(Source, self).__init__()
-        self.source = None
+        self.source_name = None
+        self.source_ra = None
+        self.source_dec = None
 
 
 class Wobble(Source):
--- a/osa/nightsummary/extract.py
+++ b/osa/nightsummary/extract.py
@@ -72,7 +72,7 @@ def extractsubruns(summary_table):
             sr.runobj.telescope = options.tel_id
             sr.runobj.night = lstdate_to_iso(options.date)
             if not options.test:
-                sr.runobj.source = database.query(
+                sr.runobj.source_name = database.query(
                     obs_id=sr.runobj.run,
                     property_name="DriveControl_SourceName"
                 )
--- a/osa/nightsummary/extract.py
+++ b/osa/nightsummary/extract.py
@@ -67,7 +67,7 @@ def extractsubruns(summary_table):
             # Build run object
             sr.runobj = RunObj()
             sr.runobj.run_str = f"{run_info['run_id']:05d}"
-            sr.runobj.run = run_info["run_id"]
+            sr.runobj.run = int(run_info["run_id"])
             sr.runobj.type = run_info["run_type"]
             sr.runobj.telescope = options.tel_id
             sr.runobj.night = lstdate_to_iso(options.date)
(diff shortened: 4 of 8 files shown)


morcuended:
Currently, we may have problems with IERS-A data from astropy (see https://github.com/astropy/astropy/issues/10494). We should download the IERS data separately (keep them up to date on a weekly basis?) and cache that data without trying to download anything when running in the cluster.

morcuended:
@moralejo @rlopezcoto @chaimain how does this scheme on post-DL2 analysis steps sounds to you?

rlopezcoto:
Sorry @morcuended , we have been very busy this week with the school. The structure looks overall good, thanks for this taking care of proposing it and for this PR. I just have a few questions:

* `std_cuts` -> have we already defined any set of std_cuts for processing? It may be a good idea to define a few different ones.
* In this structure you are introducing for the first time `source_name`, where are you expecting to get it from? 
* do we really need to include `irf_std_cuts.fits` per source? or will they be produced by lstmcpipe and you will only link them there (how are you planning to know which one?)
* I guess all this structure will be source-independent analysis, right?

chaimain:
> Sorry @morcuended , we have been very busy this week with the school. The structure looks overall good, thanks for this taking care of proposing it and for this PR. I just have a few questions:
> 
> * `std_cuts` -> have we already defined any set of std_cuts for processing? It may be a good idea to define a few different ones.
We indeed have to have a discussion on the definition of `std_cuts`, and probably store this in lstchain first.

> * In this structure you are introducing for the first time `source_name`, where are you expecting to get it from?
It seems to be by using a database query from `DriveControl_SourceName`. We should also discuss this, along with our discussion on creating and maintaining a standard observed source catalog of LST-1.

> * do we really need to include `irf_std_cuts.fits` per source? or will they be produced by lstmcpipe and you will only link them there (how are you planning to know which one?)
For IRFs, we are still awaiting the completion of various tasks - producing 'all-sky' MC list, creating the new RF model, DL2 files, merging the IRF interpolation and using energy-dependent cuts in IRF/DL3 Tools. I think it would make sense, only after these tasks, that we can talk on a 'standard' IRF type/s.

morcuended:
Hi @rlopezcoto

After talking with @chaimain I realized that there are still some open issues that are not reflected in this very simplistic scheme. The main one is the selection of proper IRFs for each observation. I'll try to answer your questions below:

- First, all this stuff refers to source-independent analysis. This is the only stream lstosa currently does. We may want to indicate this somewhere in the data tree.
- No standard cuts defined yet. Just wanted to sketch how it could be in the future. We indeed could go for several sets of cuts (e.g. tight, standard, soft cuts). This is to be discussed and agreed upon within lstchain first as @chaimain says.
- Source name associated with a given run_id (as well as source coordinates) is fetched from the TCU database in this scheme (see https://github.com/cta-observatory/lstosa/blob/main/osa/nightsummary/database.py). This is a preliminary version and needs to be discussed too. I'd say that querying this database works for runs taken from about Nov 2020 (no information is consistently there for previous runs). Also, source RADec information seems not to reflect the actual target coordinates but the coordinates plus wobbling offset. So we might want to get this drive info from drive logs instead. Although I think all this information should be centralized. I think from now on, TCU will be writing a Run Catalog having all this information. However for previously taken runs we'll have to figure out this in some other way (e.g. using https://github.com/cta-observatory/cta-lstchain/pull/880). This will need discussion among analysis, TCU and drive teams.

> - do we really need to include irf_std_cuts.fits per source? or will they be produced by lstmcpipe and you will only link them there (how are you planning to know which one?)

Here I was foreseen to have an IRF file per set of cuts not per source. This could rather be done by lstmcpipe and then we just would look for them. As @chaimain says there are several open points to be considered before we go further in lstosa.

For the moment we could produce IRF-less DL3 files and store them in the same night/date directory (without sorting them by source). We would not run either the observation indexing script for the moment. Or we could not produce DL3 files at all for the time being until previous issues are worked out.

I just wanted to move this forward so we could have automatic & fast next-day high-level results. But I guess that it only makes sense to produce theta2 & significance results for now from DL2 files or these DL3 without IRFs incorporated. Do you think this makes sense @rlopezcoto?

rlopezcoto:
thanks @morcuended and @chaimain, this sounds good for the time being, it would, however, be great if DL3 files could at least be produced for a few sets of cuts (that can be discussed as @chaimain was suggesting).

> For the moment we could produce IRF-less DL3 files and store them in the same night/date directory (without sorting them by source). We would not run either the observation indexing script for the moment. Or we could not produce DL3 files at all for the time being until previous issues are worked out.

what is the problem with running the observation indexing script?

chaimain:
> what is the problem with running the observation indexing script?
Sorry, I just checked again, and there should be no problem with running the indexing Tool for IRF-less DL3 files.

rlopezcoto:
Great, thanks guys!

chaimain:
Hi @morcuended, after discussing with @maxnoe regarding IRF-less DL3 production, we should not create DL3 files without IRFs, as it provides no additional information over the DL2 files. 
It would be better, to just create DL3 files, with the existing IRFs (Maybe point-like IRFs for now) daily with lstosa.

The lstchain DL3 Tool, will be fixed in [#709](https://github.com/cta-observatory/cta-lstchain/pull/709) to require IRFs, and take the gammaness cut information from the provided IRFs only, be it global cut or energy-dependent cuts. It will be available in the upcoming lstchain release.

So, maybe you should close #94

chaimain:
No, I have not yet discussed the "standard" cuts and type of IRFs to be produced. Will do it on Slack now, and if need be, open a github issue.

PR #89, opened 2022-01-24
Theta2 next-day results

morcuended:
This PR adds several functions to produce theta2 from DL2 files based on https://github.com/cta-observatory/2022_01_lstchain_school/blob/main/DL1b_to_Dl2/Explore_DL2_data.ipynb

Plots are produced per source for each night and the results are uploaded to the lst1 internal webpage.


morcuended:
I will merge this and test it. Probably we might want to use the post_dl2 analysis script from lstchain or even gammapy to produce the theta2 and signal significance quick results.

PR #94, opened 2022-01-31
Add option to be able to not add IRF to DL3 files

morcuended:
In case we want to add IRF to DL3 later on.


morcuended:
It turned out that the IRF-less DL3 files scheme is not viable (#81). Closing this

PR #98, opened 2022-02-01
Change name of the longterm script 

marialainez:
longterm_dl1_check.py script renamed to lstchain_longterm_dl1_check.py

### DIFF
--- a/cfg/sequencer.cfg
+++ b/cfg/sequencer.cfg
@@ -49,7 +49,7 @@ check_dl1: lstchain_check_dl1
 dl1_to_dl2: lstchain_dl1_to_dl2
 
 # Path to longterm dl1 script
-longterm_check: /fefs/aswg/software/virtual_env/ctasoft/cta-lstchain/lstchain/scripts/longterm_dl1_check.py
+longterm_check: lstchain_longterm_dl1_check
 
 # To be set by the user
 store_image_dl1ab: True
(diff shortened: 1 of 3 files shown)


morcuended:
Thanks, @marialainez! This would work. However, once lstchain PR https://github.com/cta-observatory/cta-lstchain/pull/881 is merged it would be possible to simply use the command `lstchain_longterm_dl1_check` without giving the entire path.

Then there is no need to include it in the cfg. Simply 

You have also to adapt the test to make it pass:

https://github.com/cta-observatory/lstosa/blob/0e595cec24db30d7635315813364e4204d89fad1/osa/scripts/tests/test_osa_scripts.py#L316-L328

Line 323 should simply be `"lstchain_longterm_dl1_check",` the same change goes in closer.py where this command is issued.

cfg/sequencer.cfg
```diff
@@ -49,7 +49,7 @@ check_dl1: lstchain_check_dl1
 dl1_to_dl2: lstchain_dl1_to_dl2
 
 # Path to longterm dl1 script
-longterm_check: /fefs/aswg/software/virtual_env/ctasoft/cta-lstchain/lstchain/scripts/longterm_dl1_check.py
+longterm_check: /fefs/aswg/software/virtual_env/ctasoft/cta-lstchain/lstchain/scripts/lstchain_longterm_dl1_check.py
```
morcuended, line 52: We can remove this path from here and directly use the lstchain entry point `lstchain_longterm_dl1_check` in `closer.py`

morcuended (changes requested):

morcuended (approved):

PR #104, opened 2022-02-02
Remove pedestal cleaning option in provenance tracking

Bultako:
action needed after #103

### DIFF
--- a/osa/provenance/utils.py
+++ b/osa/provenance/utils.py
@@ -124,7 +124,7 @@ def parse_variables(class_instance):
 
         class_instance.Analysisconfigfile_dl1 = os.path.realpath(configfile_dl1)
         class_instance.ObservationRun = class_instance.args[0].split(".")[0]
-        class_instance.PedestalCleaning = "True"
+        class_instance.PedestalCleaning = "False"
         class_instance.StoreImage = cfg.getboolean("lstchain", "store_image_dl1ab")
         class_instance.DL1SubrunDataset = os.path.realpath(
             f"{outdir_dl1}/dl1_LST-1.Run{class_instance.args[0]}.h5"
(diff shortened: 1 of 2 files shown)


osa/provenance/utils.py
```diff
@@ -124,7 +124,7 @@ def parse_variables(class_instance):
 
         class_instance.Analysisconfigfile_dl1 = os.path.realpath(configfile_dl1)
         class_instance.ObservationRun = class_instance.args[0].split(".")[0]
-        class_instance.PedestalCleaning = "True"
+        class_instance.PedestalCleaning = "False"
```
morcuended, line 127: Why False instead? Pedestal cleaning will be True by default

Bultako: ah sorry !
I though it was False by default.. then I see two options:
- completely remove PedestalCleaning from the provenance 
- close this PR.

morcuended: Is there a track of this config option in lstchain? If so I would say we could remove it from the osa provenance. In principle, lstosa will not set up this option from now on.

Bultako: actually there is no argument `pedestal-cleaning` aymore in `lstchain_dl1ab.py`
ok, I'm going to remove this from the provenance..

Bultako: done.

morcuended: There is no pedestal-cleaning argument anymore, but there is `--no-pedestal-cleaning` to disable it. I guess this info is kept somewhere in the metadata info of the files along with the config parameters used?

morcuended (approved):

osa/provenance/utils.py
Bultako: It does not seem to be the case, all the content of the config file is kept, which means all the values for the thresholds :(

Anyhow, I think this is something that has to be done in `lstchain_dl1ab.py`, lstosa does not set up this option.

PR #121, opened 2022-02-17
Refactor DL3+ stages

morcuended:


chaimain:
If you are adding the global theta cut, then you can specify that the gammaness cut is also global.

Also, do you intend to include energy-dependent cuts as well? They are not so well tested for classifying any 'soft', 'standard', and 'hard' cuts, but a rough estimate seems to be 80, 60, 40 %

morcuended:
> If you are adding the global theta cut, then you can specify that the gammaness cut is also global.

I'll do that, although the intention here is to have a preliminary and temporary high-level approach.

> Also, do you intend to include energy-dependent cuts as well? They are not so well tested for classifying any 'soft', 'standard', and 'hard' cuts, but a rough estimate seems to be 80, 60, 40 %

Sure it can be done. I guess the idea was to have both approaches (or only the energy-dependent one) with several sets of cuts. For the moment this is just a testbed for the processing analysis stages. I do not intend to produce any official data with it.

PR #141, opened 2022-03-12
Add script to run gain selection over a list of dates, using the "lst_select_gain" script

marialainez:


osa/scripts/gain_selection.py
```diff
@@ -0,0 +1,80 @@
...
+import click
+
+from osa.scripts.reprocessing import get_list_of_dates
+from osa.utils.logging import myLogger
+
+log = myLogger(logging.getLogger(__name__))
+
+def apply_gain_selection(date: str):
+    run_summary_file = "/fefs/aswg/data/real/monitoring/RunSummary/RunSummary_"+date+".ecsv"
+    data = ascii.read(run_summary_file)
```
morcuended, line 20: you can directly use astropy `Tables` -> `data = Tables.read(file)`

osa/scripts/gain_selection.py
```diff
@@ -0,0 +1,80 @@
...
+    data.add_index("run_id")
+    data = data[(data['run_type']=='DATA')]   # apply gain selection only to DATA runs
```
morcuended, line 22: remove redundant parenthesis

morcuended (changes requested):

osa/scripts/gain_selection.py
```diff
@@ -0,0 +1,120 @@
...
+    log.setLevel(logging.DEBUG)
+    
+    list_of_dates = get_list_of_dates(dates_file)
+
+    for date in list_of_dates:
+        apply_gain_selection(date, output_basedir)
+
+    log.info("Done! No more dates to process.")
+    
+    check_failed_jobs(output_basedir)
```
morcuended, line 116: @marialainez, in the current way of checking failed jobs, you need to launch the script again, which will submit jobs again right? Either we implement a simulate option that allows only for checking job status or the check is done in a separate script independent from the launching-job script.

PR #142, opened 2022-03-29
The datacheck directories are created and copied to the webserver even if there is no data. Create the datacheck directories only if there is data.

marialainez:


morcuended:
Thanks @marialainez. Your fix seems good to me. I just removed some functions that I implemented in the past to set no observations flags but we were not currently using. I think is better if we simply don't create the directory if not datacheck files are found.

morcuended (approved):

PR #145, opened 2022-04-23
Remove unused bash scripts

morcuended:


contrera (commented):
Please write back shortcuts.sh I use it.

morcuended:
shortcuts.sh is still in the container. I just removed it from the repository since I'm not sure it belongs here. I can bring it back if we want to back it up.

Issue #143, opened 2022-04-04
Get run date function to find calibration files is highly inefficient


morcuended:
`sequencer` currently takes several minutes to run. Below is its profile. This comes with the necessity of guessing the 'date directory' a given run was taken globing over all the R0 date subdirectories. This will get worse as the number of directories increases with time.

![image](https://user-images.githubusercontent.com/32680689/161616504-223c44c7-c91c-459d-a77d-53b47ff99246.png)

Some things that can be done:
 - Implement a `--check` option in sequencer that only returns the processing status table without having to go through all the building of sequences. Somehow detach the checking of sequences from the generation of the workflow (where the names of calibration files are defined)
 - To avoid duplication, there are already several functions in lstchain that look for calibration files/run (drs4 baseline, time, etc)

morcuended:
Since there is no run database, a possible solution would be to merge all the `RunSummary` files assigning the date to each run. The sequencer could look for the 'directory date' in this merged RunSummary file.

morcuended:
Addressed in #152

PR #152, opened 2022-05-04
Change get_run_date function to get the date corresponding to a given…

marialainez:
… run more efficiently. Solves issue #143

### DIFF
--- a/osa/configs/sequencer.cfg
+++ b/osa/configs/sequencer.cfg
@@ -29,6 +29,7 @@ DL3_DIR: %(BASE)s/DL3
 CLOSER_DIR: %(BASE)s/OSA/Closer
 HIGH_LEVEL_DIR: %(BASE)s/OSA/HighLevel
 LONGTERM_DIR: %(BASE)s/OSA/DL1DataCheck_LongTerm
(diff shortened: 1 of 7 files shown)


morcuended:
thanks @marialainez. Let me have a look

contrera:
What happens in this new scheme if the run in not in the merged_run_summaries_file.
Does it have to be included beforehand?

morcuended:
It'll fail if a run is not present in the merged summary. A cronjob will update the merged file every day with the new runs from that date. Alternatively, we can assume by default that if a run is not in the merged file, the date is the one given as input in the sequencer/closer. What do you think?

contrera:
It is fine. I would suggest to do both both:  If the run is not in the merged summary --> output a warning and use the date provided.

morcuended:
@marialainez, can you try to implement that. Something like:

```python
try:
    [current piece of code]
except IndexError:
    log.warning(
        f'Run {run_id} not found in the summary table. Assuming the date {options.date}'
    )
    date_string = options.date (in YYYYMMDD format)
```

marialainez:
yes, done in #153

PR #153, opened 2022-05-10
add a warning in case the considered run is not in the summary table

marialainez:


osa/paths.py
```diff
@@ -69,7 +69,13 @@ def get_run_date(run_id: int) -> str:
     """
     merged_run_summaries_file = cfg.get("LST1", "MERGED_SUMMARY")
     summary_table = Table.read(merged_run_summaries_file)
-    date_string = summary_table[summary_table['run_id'] == run_id]['date'][0]
+    
+    try:   
+        date_string = summary_table[summary_table["run_id"] == run_id]["date"][0]
+    except IndexError:
+        log.warning(f"Run {str(run_id)} is not in the summary table. Assuming the date of the run is {options.date}.")
+        date_string = options.date
```
morcuended, line 77: I think that `options.date` is now a datetime object, so I'm not sure whether you can simply use the `replace` method for strings. Let's see if tests pass.

osa/paths.py
```diff
@@ -69,7 +69,13 @@ def get_run_date(run_id: int) -> str:
     """
     merged_run_summaries_file = cfg.get("LST1", "MERGED_SUMMARY")
     summary_table = Table.read(merged_run_summaries_file)
-    date_string = summary_table[summary_table['run_id'] == run_id]['date'][0]
+    
+    try:   
+        date_string = summary_table[summary_table["run_id"] == run_id]["date"][0]
+    except IndexError:
+        log.warning(f"Run {str(run_id)} is not in the summary table. Assuming the date of the run is {options.date}.")
```
morcuended, line 76: I think run_id does not need to be converted to str here. The f"{run_id}" already does it.

PR #154, opened 2022-05-10
Problem reading the config file

marialainez:
If the config file is not the last argument, the read_config function is always taking as config file the DEFAULT_CFG. I add a "break" so that if a config file is passed as an argument it stops looping over other arguments that would overwrite the provided config file with the DEFAULT_CFG.


morcuended:
good catch! let me check why simulate processing test is failing

morcuended (approved):

PR #180, opened 2022-06-02
Copy the R0 files of subruns with less than 4 streams

marialainez:


osa/scripts/gain_selection.py
```diff
@@ -56,21 +57,32 @@ def apply_gain_selection(date: str, output_basedir: Path = None):
...
-            input_files = r0_dir.glob(f"LST-1.1.Run{run_id:05d}.????.fits.fz")
+            for subrun in range(n_subruns):
+                new_files = glob.glob(f"{r0_dir}/LST-1.?.Run{run_id:05d}.{subrun:04d}.fits.fz")
+
+                if len(new_files) != 4:
+                    for file in new_files:
+                        shutil.copy2(file, output_dir, follow_symlinks=True)
+ 
+                else:
+                    [input_files.append(new_files[i]) for i in range(len(new_files))]
```
morcuended, line 85: Do you have to run the gain selection script separately for each of the 4 stream files of a given subrun or would it be enough to store in `input_files` the first stream file `LST-1.1.Run{run_id:05d}.{subrun:04d}.fits.fz` once you know that the 4 streams are available?

morcuended: in any case why not do simply `input_files.extend(new_files)`?

marialainez: No, in fact providing just the first stream file is enough, I will change it

morcuended (approved):

PR #189, opened 2022-06-20
Make a hard link instead of copying files

marialainez:


osa/scripts/gain_selection.py
```diff
@@ -80,7 +79,7 @@ def apply_gain_selection(date: str, output_basedir: Path = None):
                 if len(new_files) != 4:
                     for file in new_files:
                         output_file = output_dir / file.name
-                        os.link(file, output_file, follow_symlinks=True)
+                        file.hardlink_to(output_file)
```
morcuended, line 82: I just realized that `hardlink_to` will be available as of python 3.10 (we are forced to use 3.8 at most). There is though `link_to` which will be eventually deprecated by `hardlink_to`. You can make the hard link with `file.link_to(output_file)` instead.

morcuended (approved):
Sorry for going back and forth. Should be fine now

PR #198, opened 2022-09-23
change check_failed_jobs function to catch all the jobs that did not finish in COMPLETED state

marialainez:


osa/scripts/gain_selection.py
```diff
@@ -123,27 +125,47 @@ def apply_gain_selection(date: str, output_basedir: Path = None):
         
         for file in r0_files:
             output_file = output_dir / file.name
-            file.link_to(output_file)
+            sp.run(["cp", file, output_dir])
+
+def run_sacct_j(job) -> StringIO:
```
morcuended, line 130: There should be a similar function in `osa.job`. Also, you'll probably have to take into account the `starttime` option in `sacct` since by default it only retrieves the information from the last 24 hours or even from the current day.

morcuended: The idea is to call first `run_sacct` and use its output in `get_sacct_output`. Perhaps this could be better within a unique class. But that is a different matter.

osa/scripts/gain_selection.py
```diff
@@ -123,27 +125,47 @@ def apply_gain_selection(date: str, output_basedir: Path = None):
...
+    sacct_cmd = [
+        "sacct",
+        "-n",
+        "--parsable2",
+        "--delimiter=,",
+        "--units=G",
+        "-o",
+        ",".join(FORMAT_SLURM),
+        "-j",
+        job,
```
morcuended, line 145: Now I see that here you are passing the job id. I guess then that `starttime` is not needed?

osa/scripts/gain_selection.py
```diff
@@ -123,27 +125,47 @@ def apply_gain_selection(date: str, output_basedir: Path = None):
...
 
-    for line in fileinput.input(filenames):
-        if re.search("FAILED", line) or re.search("Stream [1-4] not found", line):
-            job_id = str(fileinput.filename())[-12:-4]
-            run_id = str(fileinput.filename())[-23:-18]
-            subrun_id = str(fileinput.filename())[-17:-13]
-            failed_jobs.append(job_id)
-
-            log.warning(f"Job {job_id} (corresponding to run {run_id}, subrun {subrun_id}) failed.")
+    for job in jobs:
```
morcuended, line 157: I do not know if it's better to loop over the jobs or simply make a single sacct call based on time. However, you would then need to filter the rows corresponding to `jobs`. How long does it take to run it with the loop over jobs?

morcuended:
Also, please resolve the conflict that arose.

osa/scripts/gain_selection.py
```diff
@@ -123,27 +125,47 @@ def apply_gain_selection(date: str, output_basedir: Path = None):
...
+        return StringIO()
+
+    sacct_cmd = [
+        "sacct",
+        "-n",
+        "--parsable2",
+        "--delimiter=,",
+        "--units=G",
+        "-o",
+        ",".join(FORMAT_SLURM),
```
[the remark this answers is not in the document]
marialainez: Yes, I will do it

osa/scripts/gain_selection.py
marialainez: I also use the function get_sacct_output below, and it is not necessary to give the `starttime` option because I set the job_id.

osa/scripts/gain_selection.py
marialainez: Yes, it is not needed as I am passing the job id

osa/scripts/gain_selection.py
marialainez: It takes approximately 3 minutes per night... I will check if it's faster doing it with the `starttime` option

morcuended: Probably not. Do not bother.

osa/scripts/gain_selection.py
morcuended: Got it

morcuended (approved):

Issue #200, opened 2022-10-04
Do not use filter wheel 52 by default but use database information


morcuended:
If it fails catch the exception and then set position 52 in a second try

morcuended:
Done in #201

PR #201, opened 2022-10-05
Do not use as default filter wheel 52 in the calibration

marialainez:
solves issue #200

### DIFF
--- a/osa/paths.py
+++ b/osa/paths.py
@@ -102,11 +102,8 @@ def get_calibration_file(run_id: int) -> Path:
     """
     calib_dir = Path(cfg.get("LST1", "CALIB_DIR"))
     date = get_run_date(run_id)
-    file = calib_dir.glob(f"{date}/pro/calibration_filters_*.Run{run_id:05d}.0000.h5")
-    try:
-        return next(file)
-    except StopIteration:
-        return None
+    file = calib_dir / date / f"pro/calibration_filters_52.Run{run_id:05d}.0000.h5"
+    return file.resolve()
 
 
 def pedestal_ids_file_exists(run_id: int) -> bool:
@@ -125,10 +122,7 @@ def drs4_pedestal_exists(run_id: int) -> bool:
 def calibration_file_exists(run_id: int) -> bool:
     """Return true if calibration file was already produced."""
     file = get_calibration_file(run_id)
-    if file is None:
-        return False
-    else:
-        return True
+    return file.exists()
 
 
 def get_drive_file(date: str) -> Path:
--- a/osa/paths.py
+++ b/osa/paths.py
@@ -103,7 +103,10 @@ def get_calibration_file(run_id: int) -> Path:
     calib_dir = Path(cfg.get("LST1", "CALIB_DIR"))
     date = get_run_date(run_id)
     file = calib_dir.glob(f"{date}/pro/calibration_filters_*.Run{run_id:05d}.0000.h5")
-    return next(file)
+    try:
+        return next(file)
+    except StopIteration:
+        return None
 
 
 def pedestal_ids_file_exists(run_id: int) -> bool:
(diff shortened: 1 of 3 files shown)


osa/scripts/calibration_pipeline.py
```diff
@@ -59,7 +59,6 @@ def calibration_file_command(drs4_pedestal_run_id: int, pedcal_run_id: int) -> l
         f"--pedestal_run={drs4_pedestal_run_id}",
         f"--run_number={pedcal_run_id}",
         f"--base_dir={base_dir}",
-        "--filters=52",
```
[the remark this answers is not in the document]
marialainez: I think run 1614 could work for that, so I will test it with that run and try to implement it.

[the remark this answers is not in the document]
morcuended: Right, old runs are likely not included in the database

osa/scripts/calibration_pipeline.py
```diff
@@ -157,9 +156,19 @@ def calibrate_charge(
     cmd = calibration_file_command(
         drs4_pedestal_run_id=drs4_pedestal_run_id, pedcal_run_id=pedcal_run_id
     )
-    analysis_step = ChargeCalibrationStage(run=f"{pedcal_run_id:05d}", command_args=cmd)
-    analysis_step.execute()
-    return analysis_step.rc
+    
+    try:
+        analysis_step = ChargeCalibrationStage(run=f"{pedcal_run_id:05d}", command_args=cmd)
+        analysis_step.execute()
```
morcuended, line 162: I wonder how this works with the `.execute()` method, which retries n times if rc != 0. Maybe the retry with filter=52 could be done inside that method of the general `AnalysisStage`. Although I think it would work similarly to your implementation.

Also, have you checked whether we need some kind of cleanup before the retry?

osa/paths.py
```diff
@@ -102,8 +102,11 @@ def get_calibration_file(run_id: int) -> Path:
     """
     calib_dir = Path(cfg.get("LST1", "CALIB_DIR"))
     date = get_run_date(run_id)
-    file = calib_dir / date / f"pro/calibration_filters_52.Run{run_id:05d}.0000.h5"
-    return file.resolve()
+    file = calib_dir.glob(f"{date}/pro/calibration_filters_*.Run{run_id:05d}.0000.h5")
```
morcuended, line 105: For data processing purposes we always need the calibration file produced with filters 52. It might happen that other calibration files with different filters are produced, but we are not supposed to use them further in the analysis. So I think this function could remain the same

osa/paths.py
```diff
@@ -122,7 +125,10 @@ def drs4_pedestal_exists(run_id: int) -> bool:
 def calibration_file_exists(run_id: int) -> bool:
     """Return true if calibration file was already produced."""
     file = get_calibration_file(run_id)
-    return file.exists()
+    if file is None:
+        return False
+    else:
```
morcuended, line 130: same as above comment

osa/scripts/calibration_pipeline.py
marialainez: In principle I have tested it like this and it works. The `.execute() `method of the `ChargeCalibrationStage` doesn't retry it, it just raises an error, so if` rc!=0`, it will just execute the code under `except`: 
```
def execute(self):
        """Run the program and retry if it fails."""
        log.info(f"Executing {stringify(self.command_args)}")
        output = sp.run(self.command_args, stdout=sp.PIPE, stderr=sp.STDOUT, encoding="utf-8")
        self.rc = output.returncode
        self._write_checkpoint()

        # If fails, remove products from the directory for subsequent trials
        if self.rc != 0:
            self._clean_up()
            raise ValueError(f"{self.command} failed with output: \n {output.stdout}")
```

osa/paths.py
marialainez: ok, I will change it

PR #202, opened 2022-10-07
add possibility to analyze data for nights without DRS4 or PEDCALIB runs

marialainez:
instead of copying the DRS4 and PEDCALIB runs that need to be use to analyze those data into the run summary file of the day

### DIFF
--- a/osa/conftest.py
+++ b/osa/conftest.py
@@ -242,6 +242,74 @@ def run_summary_file(run_summary_dir):
 
 
 @pytest.fixture(scope="session")
+def run_summary_file_no_calib(run_summary_dir):
+
+    summary_content = dedent(
+        """\# %ECSV 1.0
+            # ---
+            # datatype:
+            # - {name: run_id, datatype: int64}
+            # - {name: n_subruns, datatype: int64}
+            # - {name: run_type, datatype: string}
+            # - {name: ucts_timestamp, datatype: int64}
+            # - {name: run_start, datatype: int64}
+            # - {name: dragon_reference_time, datatype: int64}
+            # - {name: dragon_reference_module_id, datatype: int16}
+            # - {name: dragon_reference_module_index, datatype: int16}
+            # - {name: dragon_reference_counter, datatype: uint64}
+            # - {name: dragon_reference_source, datatype: string}
+            # delimiter: ','
+            # meta: !!omap
+            # - {date: '2022-09-23'}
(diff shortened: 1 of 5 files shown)


morcuended:
Tests are strangely taking way too long. I wonder whether it is related to GitHub actions workflows or the changes in this PR. Have you tried to run them locally?

osa/nightsummary/tests/test_extract.py
```diff
@@ -28,3 +28,27 @@ def test_build_sequences(sequence_list):
...
+     run of another previous day is taken."""
+    from osa.nightsummary.extract import extract_runs
+    from osa.nightsummary.nightsummary import run_summary_table
+    from osa.configs import options
+
+    assert run_summary_file_no_calib.exists()
+    assert run_summary_file_no_calib2.exists()
+
+    date = datetime.fromisoformat("2022-09-23")
+    options.date = date
```
morcuended, line 44: There is a problem with mixing dates in `options.date` as it will be shared among all tests in the current approach. You could define `options.date = datetime.fromisoformat("2022-09-23")` and then right before ending the function, you define again `datetime.fromisoformat("2020-01-17")` so the other tests keep assuming that initial date.

In the long-term, it would be better to not depend on these globally defined variables and define them locally every time they are used.

osa/nightsummary/tests/test_extract.py
```diff
@@ -28,3 +27,29 @@ def test_build_sequences(sequence_list):
     extracted_seq_list = build_sequences(date=obs_date)
     for sequence, extracted_seq in zip(sequence_list, extracted_seq_list):
         assert sequence.run == extracted_seq.run
+
+
+def test_no_calib_found(run_summary_file_no_calib, run_summary_file_no_calib2):
```
morcuended, line 32: add `merged_run_summary`

osa/conftest.py
```diff
@@ -241,9 +241,79 @@ def run_summary_file(run_summary_dir):
...
+    summary_file = run_summary_dir / "RunSummary_20220922.ecsv"
+    summary_file.touch()
+    summary_file.write_text(summary_content)
+    return summary_file
+
+
 @pytest.fixture(scope="session")
 def merged_run_summary(base_test_dir):
-    """Mock merged run summary file for testing."""
+"""Mock merged run summary file for testing."""
```
morcuended, line 316: indent

morcuended (approved):

PR #203, opened 2022-10-19
Get the filter wheels from the CaCo database to create the name of th…

marialainez:
…e calib file


morcuended:
Thanks, @marialainez. Tests fail because they try to fetch filters from the database that is not accessible. Thus, we have to get around that which may not be trivial.

PR #211, opened 2023-02-13
Updates to lstchain v0.10.0

### LINKED ISSUE #206: Use new drive pointing log files
See https://github.com/cta-observatory/ctapipe_io_lst/pull/158

morcuended:
closes #206 

- [x] Update version of dependencies
- [x] Use new drive log file naming

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,8 +44,9 @@ jobs:
       - name: Setup Mamba
         uses: mamba-org/setup-micromamba@v1
         with:
-          environment-name: "osa-ci"
           environment-file: environment.yml
+          cache-environment: true
+          post-cleanup: 'all'
 
       - name: Install dependencies
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,10 +18,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up Python 3.8
+      - name: Set up Python 3.11
         uses: actions/setup-python@v2
         with:
-          python-version: 3.8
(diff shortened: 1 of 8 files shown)


morcuended:
We should try to use python 3.11 to take advantage of the 10% speed-up it offers over previous versions.

.github/workflows/ci.yml
```diff
@@ -18,10 +18,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Set up Python 3.8
-        uses: actions/setup-python@v4
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v2
```
morcuended, line 22: v4

setup.cfg
```diff
@@ -24,13 +24,13 @@ classifiers =
 [options]
 packages = find:
 setup_requires = setuptools_scm
-python_requires = >=3.8
+python_requires = >=3.11
```
morcuended, line 27: let's keep this the same as lstchain `python_requires = >=3.9` which means also we can install 3.11 (and I think it'll do by default)

setup.cfg
```diff
@@ -24,13 +24,13 @@ classifiers =
...
 install_requires =
-    astropy~=4.2
-    lstchain~=0.9.13
-    ctapipe~=0.12.0
-    matplotlib~=3.5
-    numpy<1.22.0a0
+    astropy~=5.0
+    lstchain==0.10.0
+    ctapipe~=0.19.2
```
morcuended, line 31: let's remove ctapipe and protobuf, which are not direct dependencies of lstosa, and they will be installed with lstchain

PR #214, opened 2023-02-15
Copy directly the files of runs that don't appear in the run summaries

marialainez:


morcuended:
For the record, there are spoiled runs that do not end up in the summary because they do not have the 4 streams (at least no stream 1, which is the one used to check the number of subruns in the run summary). These data runs are typically empty

morcuended (approved):

PR #215, opened 2023-02-15
Avoid running gain selection during night time and check slurm queue 

marialainez:
(as it is done in reprocessing.py)

### DIFF
--- a/osa/scripts/gain_selection.py
+++ b/osa/scripts/gain_selection.py
@@ -13,7 +13,7 @@ import click
 from astropy.table import Table
 from lstchain.paths import run_info_from_filename, parse_r0_filename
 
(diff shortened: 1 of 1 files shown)


morcuended:
check_job_status_and_wait function checks every 2 hours if there are less than 2500 pending jobs. These numbers are arbitrary to some extent, I came up with them after some tests. If they do not fit well with the gain selection script you can add these two arguments to the check_job_status_and_wait function and when you call it set whatever number of jobs and sleeping time works for this. Otherwise changes look good to me

morcuended (approved):

morcuended:
It was indeed approved but not merged. Sorry about that!

PR #219, opened 2023-03-13
Add missing line in the gain selection

marialainez:


morcuended (approved):
I overlook that before

PR #220, opened 2023-03-17
Reduce the maximum number of pending jobs to continue launching gain_…

marialainez:
…sel jobs. Also, maybe we could launch the gain selection script later so that it doesn't interfere with the lstosa processing? Maybe by changing the hours in the function "wait_for_daytime" of reprocessing.py?


morcuended:
> Maybe by changing the hours in the function "wait_for_daytime" of reprocessing.py?

Yes, can you also define the starting and end hours as input arguments for the `wait_for_daytime` function: `wait_for_daytime(start=8, end=18)` and then call the function with 12 as starting hour for the gain selection script

Also can you please move the wait_for_daytime to osa/utils/utils.py? I think it fits better there so we then do not have to call a function from a script, but from a module.

Issue #221, opened 2023-03-23
Allow for not using osa processing database


morcuended:
https://github.com/cta-observatory/lstosa/blob/db49416fbb65551b0bbe0bfd99f0fbad81fc2bb1/osa/osadb.py#L48

@marialainez, I think we could use the database if the file is defined in the cfg. Otherwise, skip everything related to it. Another way would be to catch the exception of not founding the database and continue otherwise.

When I implemented this, I thought about creating first the DB file if not exist. Another option would be to first create the DB manually with your user using the script `/fefs/aswg/data/real/OSA/database/create_osa_db.py` See also: https://github.com/cta-observatory/lstosa/wiki#osa-database

marialainez:
Ok, I will make a change so that it only writes in the database if it's defined in the cfg

morcuended:
done in #222

PR #222, opened 2023-03-23
Update the osa database only if the file is defined in the cfg file

marialainez:
solves issue #221

### DIFF
--- a/osa/scripts/sequencer.py
+++ b/osa/scripts/sequencer.py
@@ -76,7 +76,10 @@ def single_process(telescope):
     -------
     sequence_list : list
     """
-    osadb.start_processing(date_to_iso(options.date))
+    
+    database = cfg.get("database", "path")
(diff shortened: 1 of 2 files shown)


osa/scripts/sequencer.py
```diff
@@ -76,7 +76,10 @@ def single_process(telescope):
     -------
     sequence_list : list
     """
-    osadb.start_processing(date_to_iso(options.date))
+    
+    database = cfg.get("database", "path")
+    if database != "":
```
morcuended, line 81: I think in both cases it should work like this:
```
if database:
    osadb...
```

morcuended (changes requested):

morcuended (approved):

morcuended:
The failing test is related to sequencer_webmaker, unrelated to this change.

Issue #227, opened 2023-06-09
Add gain selection and DVR to OSA workflow


morcuended:
New additions to the gain selection and sequencer scripts to include the gain selection stage in the next-day onsite analysis pipeline. The straightforward approach is to perform the gain selection for the whole night first thing the morning after data taking (triggered by a cron job).

As a first step:
- [ ] Be able to run the gain-selection script with a date as an input argument instead of a list of dates (leave the possibility of also running the script with a list of dates).
- [ ] Add cronjob to run the gain selection script for a given date.
- [ ] Check job completion in gain selection. Use the existing `--check` option in the gain selection script. If everything finishes successfully, write a finished-flag file or write it in the osa database. This uses the SLURM's sacct output and lists the files in R0 and R0G respectively. Note: currently, `sacct` does not fetch correct information from SLURM jobs. This post-process of the gain selection should be checked by another cron job.
- [ ] sequencer should first check whether this flag file exists or not to keep going with the processing. `sequencer` should be called then repeatedly by a cronjob (right now there is only one call in the morning). To avoid submitting same jobs multiple times we should check in the osa DB whether the processing has already started or not. 
- [ ] Maybe we should set in the cfg whether we want lstosa to start from R0 or from R0G to be flexible.
- [ ] [OPTIONAL] It would be helpful to report the gain selection status in the sequencer output table as another column (but this is a second-order implementation).

Some aspects: 
- I guess we should keep writing the gain-selected files in R0G
- Should we keep track of the gain_selection/DVR tool version as it is outside lstchain?

The gain selection script should take ~10 minutes for each subrun. The problem is when there are runs that cannot be gain selected, thus should be just copied, which can considerably increase the overall run time.

**Note:** Alternatively, in the future, we might include the gain selection stage and any other DVR-related stage inside `datasequence.py` where all lstchain scripts are piped. In this way, there should be a previous analysis level defined in the sequence of analysis steps, which right now are R0 to DL1a -> DL1a to DL1b -> DL1 datacheck -> DL1b to DL2

marialainez:
Automatic gain selection is already implemented since some time ago, and there is another issue about the automatic DVR implementation (issue #298), so I close this issue.

PR #230, opened 2023-06-16
Add gain selection to OSA workflow

marialainez:
Related to #227


osa/configs/sequencer.cfg
```diff
@@ -137,6 +138,7 @@ WORKFLOWPREFIX: Workflow
 GRAPHSUFFIX: .dot
 SVGSUFFIX: .svg
 end_of_activity: NightFinished.txt
+gain_selection_chek: GainSelFinished.txt
```
morcuended, line 141: gain_selection_chek -> gain_selection_check
one of the tests is failing because of this

morcuended:
follow-up on this PR, conflicts need to be solved

moralejo:
> follow-up on this PR, conflicts need to be solved

Anyway, current gain selection program will not work with the data taken since early July in ADHF format, right?

morcuended:
> > follow-up on this PR, conflicts need to be solved
> 
> Anyway, current gain selection program will not work with the data taken since early July in ADHF format, right?

Ah, you're right! Then I guess this will have to be updated with whichever new tool we develop for this.

src/osa/scripts/sequencer.py
```diff
@@ -90,6 +92,10 @@ def single_process(telescope):
     if not options.simulate:
         os.makedirs(options.log_directory, exist_ok=True)
 
+    if not GainSel_finished():
+        log.info(f"Gain selection did not finish successfully for date {options.date}")
```
morcuended, line 96: Add something like. "Try again later, once gain selection has finished"

src/osa/configs/sequencer.cfg
```diff
@@ -28,6 +28,7 @@ HIGH_LEVEL_DIR: %(OSA_DIR)s/HighLevel
 LONGTERM_DIR: %(OSA_DIR)s/DL1DataCheck_LongTerm
 MERGED_SUMMARY: %(OSA_DIR)s/Catalog/merged_RunSummary.ecsv
 SEQUENCER_WEB_DIR: %(OSA_DIR)s/SequencerWeb
+GAIN_SELECTION_DIR: %(OSA_DIR)s/GainSel
```
morcuended, line 31: With this name, it seems like this is the output directory of the gain selection script instead of the gain selection ending flag

morcuended (approved):
looks good overall, if you have tested it also on actual data go ahead with the merging

PR #239, opened 2023-09-04
Use CALIB_PROD_ID instead of pro to search for calibration files

marialainez:
Otherwise, when trying to reprocess dates with a different version the new calibration files are not produced since the calibration files are found in the "pro" directory, and the reprocessing gives an error.


morcuended:
@marialainez please have a look at the changes. It would be good to test it in the container also

morcuended:
It seems there is a problem with DL1 datacheck stage of the datasequence: dl1 datacheck files are not produced.

marialainez:
I tested this branch in the IT container and it works as expected. Also, the DL1 datacheck files were produced without problems for the nights I tested.

morcuended:
> I tested this branch in the IT container and it works as expected. Also, the DL1 datacheck files were produced without problems for the nights I tested.

oh, really?! For me, the last stage (lstchain_dl1_check) was strangely running forever in the nodes. It was not producing the dl1_datacheck files. However, I ran exactly the same command issued manually and it worked fine. I could not find any apparent reason. If it works for you, I propose to resolve the current conflicts and merge it. Give it another try with main branch

PR #252, opened 2023-10-02
Use pyproject toml and update documentation

morcuended:
- Use only `pyproject.toml` instead of `setup.cfg` and `setup.py` as suggested in [project-template-python-pure](https://github.com/cta-observatory/project-template-python-pure)
- Moved code inside `src/`
- Update the home page of documentation and use PyData theme
- Linting using Ruff instead of pyflakes
- Fix some linting complaints
- Run tests using Python 3.9, 3.10 and 3.11
- Use just conda versions of `ctapipe_io_lst` and `lstchain` in `environment.yml`

It has no side effects on the way it was working before.


morcuended:
@marialainez please review. Even if it seems that there are many changes, it is just moving the files inside src, minor cosmetics stuff here and there, docs update, start using `pyproject.toml` for creating the python lstosa package and some updates in the CI (after some trials and errors) to test several python versions, some action version updates. 

**Important:** I realized that we need to stick with conda/mamba env creation for proper handling of dependencies at least in the CI and it is recommended also elsewhere.

The usage of lstosa remains the same.

marialainez (approved):

PR #256, opened 2023-10-17
Add start and end time of gain selection as input arguments 

marialainez:


morcuended (approved):

src/osa/scripts/gain_selection.py
```diff
@@ -41,7 +41,7 @@ def get_sbatch_script(
     )
 
 
-def apply_gain_selection(date: str, output_basedir: Path = None):
+def apply_gain_selection(date: str, output_basedir: Path = None, start: int, end: int):
```
morcuended, line 44: ruff is complaining about start and end arguments being after output_basedir which is optional

morcuended (approved):

Issue #225, opened 2023-04-26
Provenance being produced up to DL2 even if --no-dl2 option is used


morcuended:
Right now, provenance parsing triggered by autocloser is producing provenance products up to DL2 even if `--no-dl2` option is used. This is an unwanted side effect, which should be corrected and avoid the creation of the DL2 directories (only filled with provenance files) in this case.

Check the filters used in script `provprocess.py`

morcuended:
It would be good to fix this. OSA is still producing empty DL2 directories (only provenance products are copied there)

marialainez:
Fixed in #259

PR #259, opened 2023-11-27
Do not produce provenance for DL2 if --no-dl2 option is given

marialainez:
Fixes issue #225

### DIFF
--- a/src/osa/utils/cliopts.py
+++ b/src/osa/utils/cliopts.py
@@ -327,6 +327,12 @@ def provprocess_argparser():
         help="use this flag to reset session and remove log file",
     )
     parser.add_argument(
+        "--no-dl2",
+        action="store_true",
+        default=False,
+        help="Do not produce DL2 files (default False)",
+    )
+    parser.add_argument(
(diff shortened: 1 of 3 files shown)


src/osa/utils/cliopts.py
```diff
@@ -354,6 +354,7 @@ def provprocessparsing():
     options.configfile = opts.config.resolve()
     options.filter = opts.filter
     options.quit = opts.quit
+    options.no_dl2 = opts.no_dl2
```
morcuended, line 357: I think `no_dl2` is not defined in opts (from `provprocess_argparser`). There should be a `--no-dl2` option defined there as well. Although it might work if defined before by `closer` script, which it calls `provprocess`. In any case, I would add the cmd line option for provprocess and use it in its call

morcuended (changes requested):

morcuended (changes requested):
then you need to add the `--no-dl2` option to the `cmd` running `provprocess` in `extract_provenance` function of `closer.py` depending on the value of `options.no_dl2

morcuended (approved):

PR #264, opened 2023-12-13
Use first event time in update catalog script

### LINKED ISSUE #262: Wrong run_start values in source catalog for runs taken with EVBv6

morcuended:
Also some cleanup is done.

Closes #262

### DIFF
--- a/src/osa/scripts/update_source_catalog.py
+++ b/src/osa/scripts/update_source_catalog.py
@@ -5,30 +5,30 @@ from pathlib import Path
 from textwrap import dedent
 
 import click
-import re
 import pandas as pd
 from astropy import units as u
-from astropy.table import Table, join, vstack, unique
+from astropy.table import Table, join, unique, vstack
 from astropy.time import Time
 from lstchain.io.io import dl1_params_lstcam_key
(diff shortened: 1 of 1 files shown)


marialainez (commented):
Isn't the format and scale of the dragon_time the same as that of the times in the run_start column?

morcuended:
> Isn't the format and scale of the dragon_time the same as that of the times in the run_start column?

good point, I would have to check again

morcuended:
> Isn't the format and scale of the dragon_time the same as that of the times in the run_start column?
No, run_start in the summary was defined like
```
run_start = int(round(Time(f.camera_config.date, format="unix").unix_tai)) * int(1e9)
```
The proper way of getting the eventwise timestamps from `dragon_time` is using `time = (timestamp, format="unix", scale="utc")`. Then, getting the ISO format in UTC scale as `time.utc.iso` (to make sure to use UTC scale). So I think it is fine as it is now

morcuended:
To make the tests pass you need to merge the current main version to force the usage of lstchain >=v0.10.5

marialainez (approved):

Issue #263, opened 2023-12-12
Check state and exit code of jobs scheduled by autocloser


morcuended:
Right now there is no way of knowing whether jobs launched by autocloser finished OK or not, unless manually checked.

This can lead to unnoticed problems in the merging of files, as happened for 2023-12-11 (see https://github.com/cta-observatory/cta-lstchain/issues/1193). The DL1 merging job failed for one run, but the file was halfway produced.

The `update_source_catalog` script reads these DL1 merged files but does not check if the merging script finished properly.

morcuended:
Also even if there is a dependency on the long-term dl1 check script after all run-wise dl1 data check files are produced, I think sometimes, it is still executed regardless of the exit code of some of the previous jobs. Make sure that this does not happen, for example, launching these lstchain scripts using a bash script with `set -e` to capture the failing exit at any point.

PR #269, opened 2024-01-16
Check state of the jobs launched by autocloser

marialainez:
Check that all the jobs launched by autocloser finished correctly before creating the NightFinished.txt file. See issue #263

### DIFF
--- a/src/osa/job.py
+++ b/src/osa/job.py
@@ -702,6 +702,33 @@ def get_sacct_output(sacct_output: StringIO) -> pd.DataFrame:
     return sacct_output
 
 
+def get_closer_sacct_output(sacct_output) -> pd.DataFrame:
+    """
+    Fetch the information of jobs in the queue launched by AUTOCLOSER using the sacct 
+    SLURM output and store it in a pandas dataframe.
+
+    Returns
+    -------
+    queue_list: pd.DataFrame
+    """
+    sacct_output = pd.read_csv(sacct_output, names=FORMAT_SLURM)
+
+    # Keep only the jobs corresponding to AUTOCLOSER sequences
+    sacct_output = sacct_output[
+        (sacct_output["JobName"].str.contains("lstchain"))
+        | (sacct_output["JobName"].str.contains("provproces"))
+    ]
+
+    try:
+        sacct_output["JobID"] = sacct_output["JobID"].apply(lambda x: x.split("_")[0])
+        sacct_output["JobID"] = sacct_output["JobID"].str.strip(".batch").astype(int)
+
+    except AttributeError:
+        log.debug("No job info could be obtained from sacct")
+
+    return sacct_output
+
+
 def filter_jobs(job_info: pd.DataFrame, sequence_list: Iterable):
     """Filter the job info list to get the values of the jobs in the current queue."""
     sequences_info = pd.DataFrame([vars(seq) for seq in sequence_list])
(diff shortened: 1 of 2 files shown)


src/osa/scripts/closer.py
```diff
@@ -169,6 +175,20 @@ def post_process(seq_tuple):
     if not options.no_dl2:
         merge_files(seq_list, data_level="DL2")
 
+    time.sleep(300)
+
+    # Check if all jobs launched by autocloser finished correctly 
+    # before creating the NightFinished.txt file
+    n_max = 10
+    n = 0
+    while not all_closer_jobs_finished_correctly() & n <= n_max:
```
morcuended, line 184: So this way would keep autocloser running in background up to 5*10 mins, right?
Some numbers on the time each job usually takes:
Provprocess: up to ~12 min
Muon merging: up to ~1 min
DL1 merging: up to ~30 min!
DL1 datacheck merging runwise: up to ~5 min
Datacheck merging daily: ~3 min

So maybe the check can be done with less frequency than 5 min

And how would you know whether some jobs were not finished properly? Checking the autocloser logs? I would try something like alerting via email if not all closer jobs are finished correctly after a few checks. What do you think?

marialainez: Ok, I will add an email alert to be sent if the jobs don't finish correctly. So you suggest making the check for example every 2 minutes? And up to 30 minutes?

morcuended: no, the other way around, maybe every 10-15 min up to ~1h?

src/osa/scripts/closer.py
```diff
@@ -175,20 +176,24 @@ def post_process(seq_tuple):
...
             "All jobs launched by autocloser did not finished correctly yet. "
-            "Checking again in 5 minutes..."
+            "Checking again in 10 minutes..."
         )
-        time.sleep(300)
+        time.sleep(600)
         n += 1
 
+    if n > n_max:
+        send_warning_mail(date=options.date)
```
morcuended, line 194: the problem with this way of sending mail is that in the current scheme, in which autocloser is launched by a cron job twice per hour for the whole day, it will keep sending warning mails. In this sense, a unique cronjob checking the night is finished night once per day, and if not sending an email, could be a better solution.

Also since now running autocloser could take up to 1 hour, the frequency of the cron job has to be updated correspondingly.

src/osa/job.py
```diff
@@ -702,6 +702,36 @@ def get_sacct_output(sacct_output: StringIO) -> pd.DataFrame:
...
+    queue_list: pd.DataFrame
+    """
+    sacct_output = pd.read_csv(sacct_output, names=FORMAT_SLURM)
+
+    # Keep only the jobs corresponding to AUTOCLOSER sequences 
+    # Until the merging of muon files is fixed, check all jobs except "lstchain_merge_muon_files"
+    sacct_output = sacct_output[
+        (sacct_output["JobName"].str.contains("lstchain_merge_hdf5_files"))
+        | (sacct_output["JobName"].str.contains("lstchain_check_dl1"))
+        | (sacct_output["JobName"].str.contains("lstchain_longterm_dl1_check"))
```
morcuended, line 721: add the new cherenkov transparency step jobname

morcuended (approved):
I think we can go ahead with this, but keep in mind that cronjob frequency should change accordingly, with these changes the autocloser could be running up to 1 h checking for job proper completion.

PR #271, opened 2024-02-02
Launch lstchain_cherenkov_transparency.py script as a new step of closer

marialainez:


src/osa/scripts/closer.py
```diff
@@ -157,6 +158,12 @@ def post_process(seq_tuple):
         list_job_id = merge_dl1_datacheck(seq_list)
         daily_datacheck(daily_longterm_cmd(list_job_id))
 
+    # Wait 5 minutes for the longterm DL1 datacheck file to be produced
+    time.sleep(300)
+
+    # Update longterm DL1 datacheck file with cherenkov_transparency script
+    cherenkov_transparency()
```
morcuended, line 165: better to add a slurm dependency between both jobs as done between runwise dl1 checks and the daily longterm check jobs

src/osa/scripts/closer.py
```diff
@@ -157,6 +158,12 @@ def post_process(seq_tuple):
         list_job_id = merge_dl1_datacheck(seq_list)
         daily_datacheck(daily_longterm_cmd(list_job_id))
```
morcuended, line 159: so daily_datacheck function should return the job_id then pass it to `cherenkov_transparency` which should call sbatch as 
```python
        "sbatch",
        "-D",
        options.directory,
        "-o",
        "log/cherenkov_transparency_%j.log",
        f"--dependency=afterok:{longterm_job_id)}",
        "lstchain_cherenkov_transparency",
        "blabla"
```

morcuended (changes requested):
looks good, just try to implement the job dependency

src/osa/scripts/closer.py
```diff
@@ -484,6 +486,39 @@ def daily_datacheck(cmd: List[str]):
     log.info("Daily dl1 checks using longterm script.")
     log.debug(f"Executing {stringify(cmd)}")
 
+    if not options.simulate and not options.test and shutil.which("sbatch") is not None:
+        job = subprocess.run(cmd, check=True)
```
morcuended, line 490: One last thing, in order to sbatch to return the job id, you need to add `--parsable` argument after sbatch cmd in `daily_longterm_cmd` like:

```
"sbatch",
"--parsable",
"-D",
options.directory,
"-o",
...
```

morcuended (changes requested):

morcuended (approved):
I think it should work like this

morcuended:
Test is failing because the expected command in it does not contain the `--parsable`. Please add it to the expected command.

morcuended (approved):

PR #273, opened 2024-02-07
Make link of the updated longterm datacheck file in a common directory

marialainez:


src/osa/scripts/closer.py
```diff
@@ -526,5 +529,44 @@ def cherenkov_transparency(cmd: List[str]):
...
+    nightdir = date_to_dir(options.date)
+    latest_version = 0
+    for file in longterm_files:
+        idx1 = file.find("/v0.")
+        idx2 = file.find(f"/{nightdir}")
+        version = file[idx1+1:idx2]
+        if int(version[3:])>latest_version:
+            latest_version = int(version[3:])
+            
+    return "v0."+str(latest_version)
```
morcuended, line 543:
```suggestion
    latest_version = None
    for file in longterm_files:
        version = file.parents[1].name
        version_number = int(version.split(".")[1])
        if latest_version is None or version_number > latest_version:
            latest_version = version_number
    
    return f"v0.{latest_version}"
```

src/osa/scripts/closer.py
```diff
@@ -526,5 +529,44 @@ def cherenkov_transparency(cmd: List[str]):
...
+
+
+def create_longterm_symlink():
+    """If the created longterm DL1 datacheck file corresponds to the latest 
+    version available, make symlink to it in the "all" common directory."""
+    nightdir = date_to_dir(options.date)
+    longterm_dir = Path(cfg.get("LST1", "LONGTERM_DIR"))
+    longterm_datacheck_file = longterm_dir / options.prod_id / nightdir / f"DL1_datacheck_{nightdir}.h5"
+    linked_longterm_file = longterm_dir / f"night_wise/all/DL1_datacheck_{nightdir}.h5"
+    all_longterm_files = glob.glob(longterm_dir + f"/v*/{nightdir}/DL1_datacheck_{nightdir}.h5")
```
morcuended, line 553: it could also be that given the `all_longterm_files` for that given date, you can get the latest-prod-id one with:
```python
def get_latest_version(longterm_files):
    latest_version_path = max(longterm_files, key=lambda path: path.parent.name.split(".")[1] if path.parent.name.startswith("v") else "")
    return latest_version_path
```
and then always link that version without having to check if there is only 1 longterm file or several. Simply link always the last one? Would that work?

morcuended (changes requested):
I left some comments, and possible simplifications

src/osa/scripts/closer.py
```diff
@@ -526,5 +529,44 @@ def cherenkov_transparency(cmd: List[str]):
...
+    return "v0."+str(latest_version)
+
+
+def create_longterm_symlink():
+    """If the created longterm DL1 datacheck file corresponds to the latest 
+    version available, make symlink to it in the "all" common directory."""
+    nightdir = date_to_dir(options.date)
+    longterm_dir = Path(cfg.get("LST1", "LONGTERM_DIR"))
+    longterm_datacheck_file = longterm_dir / options.prod_id / nightdir / f"DL1_datacheck_{nightdir}.h5"
+    linked_longterm_file = longterm_dir / f"night_wise/all/DL1_datacheck_{nightdir}.h5"
```
morcuended, line 552: you might need to ensure that the directory with linked files `longterm_dir / f"night_wise/all` exists here

src/osa/scripts/closer.py
marialainez: yes, I think it works, but I needed to make some changes to that function to work correctlty

src/osa/paths.py
```diff
@@ -141,6 +142,7 @@ def get_calibration_filename(run_id: int, prod_id: str) -> Path:
         mongodb = cfg.get("database", "caco_db")
         try:
             # Cast run_id to int to avoid problems with numpy int64 encoding in MongoDB
+            #options.filters = find_filter_wheels(int(run_id), mongodb)
```
morcuended, line 145:
```suggestion

```

src/osa/scripts/tests/test_osa_scripts.py
```diff
@@ -198,6 +199,7 @@ def test_closer(
     for obs_file in test_observed_data:
         assert obs_file.exists()
     assert merged_run_summary.exists()
+    assert longterm_dir.exists()
```
morcuended, line 202: on top of longterm dir you also should create before the dir with all links. I suggest creating:

```python
@pytest.fixture(scope="session")
def longterm_link_latest_dir(base_test_dir):
    directory = base_test_dir / "OSA" / "DL1DataCheck_LongTerm" / to / expected / location
    directory.mkdir(parents=True, exist_ok=True)
    return directory
```
then also assert its existence in test_closer

src/osa/paths.py
```diff
@@ -346,3 +347,23 @@ def create_source_directories(source_list: list, cuts_dir: Path):
...
+    """If the created longterm DL1 datacheck file corresponds to the latest 
+    version available, make symlink to it in the "all" common directory."""
+    nightdir = date_to_dir(options.date)
+    longterm_dir = Path(cfg.get("LST1", "LONGTERM_DIR"))
+    linked_longterm_file = longterm_dir / f"night_wise/all/DL1_datacheck_{nightdir}.h5"
+    all_longterm_files = sorted(longterm_dir.rglob(f"v*/{nightdir}/DL1_datacheck_{nightdir}.h5"))
+    latest_version_file = get_latest_version_file(all_longterm_files)
+
+    log.info("Make symlink of the latest version longterm DL1 datacheck file in the common directory.")
+    linked_longterm_file.unlink()
```
morcuended, line 368:
```suggestion
    linked_longterm_file.unlink(missing_ok=True)
```
I think this line for removing the existing link can be removed. Simply create the link, if it already exists, it will be overwritten.

>  If the target is present, the type of the symlink will be created to match. Otherwise, the symlink will be created as a directory if target_is_directory is True or a file symlink (the default) otherwise.

src/osa/scripts/tests/test_osa_scripts.py
```diff
@@ -179,6 +179,8 @@ def test_closer(
     drs4_time_calibration_files,
     systematic_correction_files,
     merged_run_summary,
+    longterm_dir,
+    longterm_link_latest_dir,
```
morcuended, line 183:
```suggestion
    longterm_link_latest_dir,
    daily_datacheck_dl1_files,
```

src/osa/scripts/tests/test_osa_scripts.py
```diff
@@ -198,6 +200,8 @@ def test_closer(
     for obs_file in test_observed_data:
         assert obs_file.exists()
     assert merged_run_summary.exists()
+    assert longterm_dir.exists()
+    assert longterm_link_latest_dir.exists()
```
morcuended, line 204:
```suggestion
    assert longterm_link_latest_dir.exists()
    for check_file in daily_datacheck_dl1_files:
        assert check_file.exists()
```

morcuended (changes requested):

morcuended (approved):
go ahead with the merging

PR #278, opened 2024-02-14
Undo the possible symlink to longterm file before making the new one

marialainez:
After testing lstosa v0.10.9 with a night that was already processed with v0.9, I got an error due to an already existent link to the longterm file in the global directory "all", so I add a line to unlink this file before making the new symlink.

### DIFF
--- a/src/osa/paths.py
+++ b/src/osa/paths.py
@@ -369,4 +369,5 @@ def create_longterm_symlink():
     latest_version_file = get_latest_version_file(all_longterm_files)
 
     log.info("Symlink the latest version longterm DL1 datacheck file in the common directory.")
+    linked_longterm_file.unlink()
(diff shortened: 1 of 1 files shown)


src/osa/paths.py
```diff
@@ -369,4 +369,5 @@ def create_longterm_symlink():
     latest_version_file = get_latest_version_file(all_longterm_files)
 
     log.info("Symlink the latest version longterm DL1 datacheck file in the common directory.")
+    linked_longterm_file.unlink()
```
morcuended, line 372:
```suggestion
    linked_longterm_file.unlink(missing_ok=True)
```

morcuended (changes requested):
In the case the file does not exist

morcuended (approved):

PR #279, opened 2024-02-21
Wait until the longterm datacheck file is created to make the link

marialainez:
Currently, autocloser is trying to make the symlink of the file right after the cherenkov_transparency script is launched, and it fails because the file has not yet been created. I put long times to wait for the job completion because nowadays there are many jobs running in the cluster, and many of them stay in PENDING status for a long time until they start running. 

EDIT:

> The first time closer runs it will go through the else, and wait to see if all the jobs are finished (until the hour), if so, the flag is created. If it is not closed after that hour, there is no nightfinished flag. Then closer runs again, first it checks that the longterm file exists, if it does it goes to the job completion check directly and tries to close again.

### DIFF
--- a/src/osa/paths.py
+++ b/src/osa/paths.py
@@ -362,7 +362,11 @@ def get_latest_version_file(longterm_files: List[str]) -> Path:
 
 
 def is_job_completed(job_id: str):
-    """Wait until the SLURM job corresponding to job_id finishes."""
+    """
+    Check whether SLURM job `job_id` has finished.
+    
+    It keeps checking every 10 minutes for one our.
+    """
     n_max = 10
     n = 0
     while n < n_max:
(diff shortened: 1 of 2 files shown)


src/osa/paths.py
```diff
@@ -359,15 +361,35 @@ def get_latest_version_file(longterm_files: List[str]) -> Path:
     )
 
 
-def create_longterm_symlink():
+def wait_for_job_completion(job_id: str):
```
morcuended, line 364: maybe beter name `is_job_completed`

that later reads

```python
if is_job_completed(job_id):
   do the link
else:
   cannot do the link, try again later
```

src/osa/scripts/closer.py
```diff
@@ -154,32 +158,38 @@ def ask_for_closing():
 def post_process(seq_tuple):
     """Set of last instructions."""
     seq_list = seq_tuple[1]
+    
+    if dl1_datacheck_longterm_file_exits() and not options.test:
+        create_longterm_symlink()
```
morcuended, line 163: I think this can be checked at the very beginning of the closer main, not here in `post_process`

morcuended (changes requested):

src/osa/scripts/closer.py
morcuended: Right, right, did not realize that.

PR #280, opened 2024-02-28
Look for the closest declination line and adequate NSB random forest model for DL2 production

### LINKED ISSUE #275: DL2 production with standard dark NSB models
There is interest in resuming the DL2 production using the all-sky dark NSB standard MC. This will speed up the preliminary next-day analysis.

As an example, this is a piece of code that looks for the closest declination line. Should be integrated in lstosa

```python
def convert_dec_string(dec_str):
    # Split the string into parts
    parts = dec_str.split('_')

    # Extract the sign, degrees, and minutes
    sign = 1 if 'min' not in parts else -1
    degrees = int(parts[-1])

    # Calculate the numerical value
    dec_value = sign * (degrees / 100)

    return dec_value  # deg


def get_corresponding_string(list1, list2):
    corresponding_dict = {}
    for index, element in enumerate(list2):
        corresponding_dict[element] = list1[index]
    return corresponding_dict


@click.command()
@click.option("--target-name", type=str, required=True)
def main(target_name: str):    
    source_coordinates = SkyCoord.from_name(target_name)
    
    dec_list = os.listdir("/fefs/aswg/data/mc/DL0/LSTProd2/TrainingDataset/Protons")
    config = paths_config.PathConfigAllSkyFull('all_decs', dec_list)
    
    
    # Convert each string in the list to numerical values
    dec_values = [convert_dec_string(dec) for dec in dec_list]
    
    # Print the resulting numerical values
    print("List of declinations:")
    for dec_str, dec_value in zip(dec_list, dec_values):
        print(f'{dec_str}: {dec_value}')

    print("\n")
    
    closest_declination = min(dec_values, key=lambda x: abs(x - source_coordinates.dec.value))
    
    print(f"The declination closest to {source_coordinates.dec.value:.2f} is: {closest_declination}")
    
    corresponding_dict = get_corresponding_string(dec_list, dec_values)
    corresponding_string = corresponding_dict[closest_declination]
    
    list_nodes = sorted(BASE_MODELS.rglob(corresponding_string), key=os.path.getmtime)

    print("\nPaths:")
    for path in list_nodes:
        print(path)
```

marialainez:
Choose the adequate RF model in the [declination lines x NSB] MC grid.

Closes #275  (initially for just choosing the declination MC track closest to the target declination)


src/osa/paths.py
contrera, line 1: Is it possible to have the basic paths used in the functions, like  BASE_MODELS = Path("/fefs/aswg/data/models/AllSky") or run_catalog_dir = Path("/fefs/aswg/data/real/monitoring/RunCatalog")  through cfg calls as it was done previously with rf_models_directory = Path(cfg.get("lstchain", "RF_MODELS"))

src/osa/paths.py
```diff
@@ -397,10 +399,100 @@ def create_longterm_symlink(cherenkov_job_id: str = None):
...
+    corresponding_dict = {}
+    for index, element in enumerate(list2):
+        corresponding_dict[element] = list1[index]
+    return corresponding_dict
+
+
+def get_latest_RF_model_path(dec_str: str) -> Path:
+    """Get the path of the most recent version of RF models for a given declination,
+    excluding the ones produced for the source-dependent analysis."""
+    BASE_MODELS = Path("/fefs/aswg/data/models/AllSky")
```
morcuended, line 438: define this in the cfg

src/osa/paths.py
```diff
@@ -397,10 +399,100 @@ def create_longterm_symlink(cherenkov_job_id: str = None):
...
+    # make sure the RF models correspond to the current version of lstchain
+    current_version = get_major_version(utils.get_lstchain_version())
+    list_nodes = sorted(BASE_MODELS.rglob(f"*{current_version}*/{dec_str}"), key=os.path.getmtime)
```
morcuended, line 441: maybe also the all-sky MC production could be defined in the cfg as well for a given lstchain major release. So no need to glob over them

src/osa/paths.py
```diff
@@ -397,10 +399,100 @@ def create_longterm_symlink(cherenkov_job_id: str = None):
...
+    for i in list_nodes:
+        if "srcdep" in str(i):
+            list_nodes.remove(i)
+
+    return list_nodes[-1]
+        
+
+def get_RF_model(run_str: str) -> Path:
+    """Get the path of the RF model to be used in the DL2 production for a given run."""
+    run_catalog_dir = Path("/fefs/aswg/data/real/monitoring/RunCatalog")
```
morcuended, line 457: use cfg definition

src/osa/paths.py
```diff
@@ -397,10 +399,100 @@ def create_longterm_symlink(cherenkov_job_id: str = None):
...
+    run_catalog = Table.read(run_catalog_file)
+    run = run_catalog[run_catalog["run_id"]==int(run_str)]
+    
+    try:
+        target_name = run["source_name"]
+        source_coordinates = SkyCoord.from_name(target_name)
+        source_dec = source_coordinates.dec.value
+        
+    except TypeError:
+        source_dec = run["source_dec"][0]
```
morcuended, line 468: In the way it is fetched right now from the TCU database, this is not exactly the declination of the source, but the declination at which the telescope is "pointing". It should be possible to look for the target coordinates in the TCU database though.

src/osa/paths.py
```diff
@@ -397,10 +399,100 @@ def create_longterm_symlink(cherenkov_job_id: str = None):
...
+    list_nodes = sorted(BASE_MODELS.rglob(f"*{current_version}*/{dec_str}"), key=os.path.getmtime)
+    
+    log.debug(f"Found len(list_nodes) paths with {current_version} corresponding to {dec_str}:")
+    for path in list_nodes:
+        log.debug(path)
+    
+    # remove from the list the models produced for the source-dependent analysis        
+    for i in list_nodes:
+        if "srcdep" in str(i):
+            list_nodes.remove(i)
```
morcuended, line 450: See above. Defining directly the MC prod (after `...models/AllSky/`) would avoid this

morcuended (changes requested):

morcuended:
@moralejo, I was going to comment in today's meeting. The way we look for the adequate RF model declination track is by using the pointing information from the TCU database (specifically the target coordinates), not checking the average pointing information inside the data files as you mentioned. Is this fine?

moralejo:
Since the relevant pointing is the one in celestial coordinates, the target coordinates are perfectly fine for standard observations (if we ever do something like drift scans we will anyway have to do some custom work).

src/osa/paths.py
```diff
@@ -397,10 +399,85 @@ def create_longterm_symlink(cherenkov_job_id: str = None):
...
+def get_RF_model(run_str: str) -> Path:
+    """Get the path of the RF model to be used in the DL2 production for a given run."""
+    run_catalog_dir = Path(cfg.get(options.tel_id, "RUN_CATALOG"))
+    run_catalog_file = run_catalog_dir / f"RunCatalog_{utils.date_to_dir(options.date)}.ecsv"
+    run_catalog = Table.read(run_catalog_file)
+    run = run_catalog[run_catalog["run_id"]==int(run_str)]
+    target_name = run["source_name"]
+
+    try:
+        source_coordinates = SkyCoord.from_name(target_name[0])
```
morcuended, line 444: as checked by @marialainez, this is not working when running the job in batch using SLURM, since the SLURM nodes are not connected to the outside world via internet

morcuended (commented):
Looks good! thanks @marialainez. I leave some comments. I'd like to test it a bit myself. How about the memory and time requirements of the jobs including the dl1 to dl2 stage?

src/osa/paths.py
```diff
@@ -7,9 +7,11 @@
 from typing import List
 import subprocess
 import time
+import os
 
 import lstchain
 from astropy.table import Table
+from astropy.coordinates import SkyCoord
```
morcuended, line 14: remove unused import

morcuended (changes requested):

morcuended (commented):
Let's try to push this forward. I left some comments. I think the tricky part is to readjust the job's requirements (time and memory) when including dl1 to dl2 step, although that is independent from this PR.

src/osa/paths.py
```diff
@@ -397,10 +399,85 @@ def create_longterm_symlink(cherenkov_job_id: str = None):
...
     longterm_file = longterm_dir / options.prod_id / nightdir / f"DL1_datacheck_{nightdir}.h5"
     return longterm_file.exists()
 
+
+def convert_dec_string(dec_str: str) -> float:
+    """Return the declination angle in degrees corresponding to a 
+    given string of the form "dec_XXXX" or "dec_min_XXXX"."""
+    
+    # Check if dec_str has a valid format
+    pattern = r'^dec(_min)?_[0-9]{4}$'
```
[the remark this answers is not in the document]
marialainez: done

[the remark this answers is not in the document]
morcuended: I do not see it changed yet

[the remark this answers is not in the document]
marialainez: Sorry, I thought I had already committed it. Now it's done

marialainez:
I tested the dl2 production with this branch and it works correctly with MEMSIZE_DATA: 60 GB.

src/osa/utils/utils.py
```diff
@@ -285,3 +291,118 @@ def wait_for_daytime(start=8, end=18):
...
+
+        return dec_value*u.deg
+
+
+def get_corresponding_string(list1: list, list2: list) -> dict:
+    """Return a dictionary created from two given lists."""
+    corresponding_dict = {}
+    for index, element in enumerate(list2):
+        corresponding_dict[element] = list1[index]
+    return corresponding_dict
```
morcuended, line 348: Try to name this function with a more descriptive name.

src/osa/utils/utils.py
```diff
@@ -285,3 +291,118 @@ def wait_for_daytime(start=8, end=18):
...
+            declination_str = corresponding_dict[closest_declination]
+            dec_values.remove(closest_declination)
+            dec_list.remove(declination_str)
+            closest_declination = min(dec_values, key=lambda x: abs(x - source_dec))
+            closest_dec_culmination = culmination_angle(closest_declination)
+    
+    log.debug(f"The declination line to use for the DL2 production is: {closest_declination}")
+    
+    corresponding_dict = get_corresponding_string(dec_list, dec_values)
+    declination_str = corresponding_dict[closest_declination]
```
morcuended, line 404: Also here try to use more descriptive names for these variables, so it is clearer when going through the code.

morcuended (changes requested):
A couple of additional comments. Otherwise looks good.

morcuended (dismissed):

src/osa/utils/utils.py
```diff
@@ -285,3 +292,137 @@ def wait_for_daytime(start=8, end=18):
...
+    }
+    return nsb_dict
+
+
+def get_mc_nsb_dir(run_id: int, rf_models_dir: Path) -> Path:
+    """
+    Return the path of the RF models directory with the NSB level 
+    closest to that of the data for a given run.
+    """
+    input_dir = Path(options.directory)
```
morcuended, line 366: maybe use another name for `input_dir`, like `analysis_dir` or `dl1_dir` to explicitly say that this are the input files that you use for tailcut calculation

src/osa/utils/utils.py
```diff
@@ -285,3 +292,137 @@ def wait_for_daytime(start=8, end=18):
...
+    # Convert each string in the list to numerical values
+    dec_values = [convert_dec_string(dec) for dec in dec_list]
+    dec_values = [dec for dec in dec_values if dec is not None]
+    
+    closest_declination = min(dec_values, key=lambda x: abs(x - source_dec))
+    closest_dec_culmination = culmination_angle(closest_declination)
+    
+    location = observatory_locations["cta_north"]
+    Lat = location.lat  # latitude of the LST1 site    
+    closest_lines = sorted(sorted(dec_values, key=lambda x: abs(x - Lat))[:2])
```
morcuended, line 402:
```suggestion
    lst_location = observatory_locations["cta_north"]
    lst_latitude = lst_location.lat  # latitude of the LST1 site    
    closest_lines = sorted(sorted(dec_values, key=lambda x: abs(x - lst_latitude))[:2])
```
more explicit names for ease the reading of the code

src/osa/utils/utils.py
```diff
@@ -285,3 +292,137 @@ def wait_for_daytime(start=8, end=18):
...
+
+    return nsb_dict[closest_nsb_value]
+
+
+def get_RF_model(run_id: int) -> Path:
+    """Get the path of the RF model to be used in the DL2 production for a given run."""
+    datacheck_dir = options.directory / options.dl1_prod_id / "datacheck"
+    datacheck_file = datacheck_dir / f"datacheck_dl1_LST-1.Run{run_id:05d}.h5"
+
+    source_dec = get_median_dec(datacheck_file)
```
morcuended, line 381: this is no longer source declination but *telescope pointing* declination. I would rename these variables to avoid confusion in the future.

src/osa/utils/utils.py
```diff
@@ -285,3 +292,137 @@ def wait_for_daytime(start=8, end=18):
...
+
+    rf_models_prefix = cfg.get("lstchain", "mc_prod")
+    nsb_dict = get_nsb_dict(rf_models_dir, rf_models_prefix)
+    closest_nsb_value = min(nsb_dict.keys(), key=lambda x: abs(float(x) - additional_nsb))
+
+    return nsb_dict[closest_nsb_value]
+
+
+def get_RF_model(run_id: int) -> Path:
+    """Get the path of the RF model to be used in the DL2 production for a given run."""
```
morcuended, line 377:
```suggestion
    """Get the path of the RF models to be used in the DL2 production for a given run.
    
    The choice of the models is based on the adequate additional NSB level
    and the proper declination line of the MC used for the training.
    """
```

morcuended (changes requested):
The NSB + declination directory scheme looks good to me. Just left some comments to improve the readability of the code.

I trigger again the CI, since now lstchain is available in conda forge.

EDIT on the CI run: it will not work because of an incompatibility with lstcam_calib package still requiring ctapipe_io_lst 0.24.x

marialainez:
I did that because the dl2 files should be saved in directories like "...tailcutXY/nsb_tuning_X.XX/", and the cleaning levels are fixed in the dl1ab step with the find_tailcuts function.

marialainez:
Using as pointing declination the median of the declination of the datacheck files was not working, because by the time the sequence*.py files are written, the datacheck files have not been produced yet. Therefore, the RF model could not be determined. So I made this change to use the declination given in the run catalogs, which is actually the pointing declination of the telescope.

marialainez:
No, in any case, I am testing it and this scheme doesn't work, because the function get_RF_model() is called before having even launched the r0_to_dl1 step. Consequently, no dl1a files have been created yet by the time this function is called and therefore the "find_tailcuts" function inside get_RF_model cannot be called. 
I am thinking on going back to the initial scheme and calling to the get_RF_model function inside datasequence in the dl1_to_dl2 step. In this way, we could still use the median declination from the datacheck files. 
The reason why we moved to this new scheme was to avoid accessing to the TCU database for every subrun, but we do not get the declination from there anymore.

morcuended:
I agree, it seems like another analysis step that could fit in datasequence.

rcervinoucm (approved):

Issue #281, opened 2024-03-04
Remove broken and deprecated show_run_summary script


morcuended:
Right now, the script the shifters use to see the observed time is `show_run_summary_tcu`. Previously we used `show_run_summary`, which did not depend on TCU info but on looking at the R0 files directly. This last script is not compatible with the latest version of lstchain that is being used:

```python
Traceback (most recent call last):
  File "/fefs/aswg/software/conda/envs/osa-v0.10/bin/show_run_summary", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/fefs/aswg/software/conda/envs/osa-v0.10/lib/python3.11/site-packages/osa/scripts/show_run_summary.py", line 128, in main
    reference_counters = [read_counters(date_path, run) for run in run_numbers]
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fefs/aswg/software/conda/envs/osa-v0.10/lib/python3.11/site-packages/osa/scripts/show_run_summary.py", line 128, in <listcomp>
    reference_counters = [read_counters(date_path, run) for run in run_numbers]
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: read_counters() takes 1 positional argument but 2 were given
```
`read_counters` comes from `lstchain.scripts.lstchain_create_run_summary`.

I propose simply removing the `show_run_summary` script since anyway it hasn't been used for some time and we would be maintaining two scripts. Then I would rename `show_run_summary_tcu` as `show_run_summary` (it also has to be changed in the launcher script that shifters use)

PR #283, opened 2024-03-05
Remove show_run_summary + rename show_run_summary_tcu

marialainez:
solves issue #281


morcuended (approved):
thanks! once the new lstosa release is in production do not forget to change the way shifters call the show run summary script `show_run_summary_tcu` -> `show_run_summary`

Issue #287, opened 2024-05-01
sequencer webmaker reports no observations when actually there are


morcuended:
happening in the latest lstosa release, might be related to new no gain sel option in sequencer (which is run inside the sequencer webmaker script)

https://github.com/cta-observatory/lstosa/blob/962da4d31c397298f8eb61adb3faf8a544d5f0b1/src/osa/scripts/sequencer_webmaker.py#L74-L83

should include no-gain-sel option otherwise will not return the correct output.

```
(osa-v0.10) [lstanalyzer@tcs06 ~]$ sequencer -c /fefs/aswg/lstosa/cfg/sequencer_v0.10.cfg -s LST1
=================================== Starting sequencer.py at 2024-05-01 12:02 UTC for LST, Telescope: LST1, Date: 2024-04-30 ===================================
Gain selection did not finish successfully for date 2024-04-30 12:02:09.524403.Try again later, once gain selection has finished.
```

while:
```
(osa-v0.10) [lstanalyzer@tcs06 ~]$ sequencer -c /fefs/aswg/lstosa/cfg/sequencer_v0.10.cfg -s --no-gainsel LST1
=================================== Starting sequencer.py at 2024-05-01 12:04 UTC for LST, Telescope: LST1, Date: 2024-04-30 ===================================
Tel   Seq  Parent  Type      Run    Subruns  Source  Action  Tries  JobID     State      CPU_time  Exit  DL1%  MUONS%  DL1AB%  DATACHECK%  DL2%  
LST1    1  None    PEDCALIB  17208  2        None    Closed      1  34739565  COMPLETED  00:08:41  0:0   None  None    None    None        None  
...
```

morcuended:
solved in #288

PR #288, opened 2024-05-02
Add no-gainsel option to sequencer_webmaker

marialainez:
It should solve issue #287 . The --no-gainsel option should be added to the crontab in the cluster.


morcuended (approved):
thx, looks good

PR #292, opened 2024-05-28
Add Cat-B calibration to OSA

marialainez:


src/osa/configs/sequencer.cfg
```diff
@@ -15,9 +15,10 @@ RUN_SUMMARY_DIR: %(MONITORING)s/RunSummary
 RUN_CATALOG: %(MONITORING)s/RunCatalog
 PEDESTAL_FINDER_DIR: %(BASE)s/auxiliary/PedestalFinder
 ANALYSIS_DIR: %(BASE)s/running_analysis
-CALIB_BASE_DIR: %(MONITORING)s/PixelCalibration/Cat-A
-CALIB_DIR: %(CALIB_BASE_DIR)s/calibration
-PEDESTAL_DIR: %(CALIB_BASE_DIR)s/drs4_baseline
+CALIB_BASE_DIR: %(MONITORING)s/PixelCalibration
+CALIB_DIR: %(CALIB_BASE_DIR)s/Cat-A/calibration
+PEDESTAL_DIR: %(CALIB_BASE_DIR)s/Cat-A/drs4_baseline
+CATB_CALIB_DIR: %(CALIB_BASE_DIR)s/Cat-B
```
morcuended, line 21:
```suggestion
CALIB_BASE_DIR: %(MONITORING)s/PixelCalibration
CAT_A_CALIB_BASE: %(CALIB_BASE_DIR)s/Cat-A
CAT_A_CALIB_DIR: %(CAT_A_CALIB_BASE)s/calibration
CAT_A_PEDESTAL_DIR: %(CAT_A_CALIB_BASE)s/drs4_baseline
CAT_B_CALIB_BASE: %(CALIB_BASE_DIR)s/Cat-B
```

morcuended: I would specify each Cat A and Cat B name in the directories. It has to be change in the code accordingly

src/osa/scripts/datasequence.py
```diff
@@ -189,6 +231,10 @@ def dl1ab(run_str: str) -> int:
     input_dl1_datafile = Path(options.directory) / f"dl1_LST-1.Run{run_str}.h5"
     # DL1b output file to be stored in the dl1ab subdirectory
     output_dl1_datafile = dl1ab_subdirectory / f"dl1_LST-1.Run{run_str}.h5"
+    night_dir = date_to_dir(options.date)
+    calib_prod_id = cfg.get("LST1", "CALIB_PROD_ID")
```
morcuended, line 235: maybe this calib prod id could be simply the lstchain major version? v0.10, v0.11, etc

morcuended (commented):
Looks good overall, I left a few first set of comments. Thanks also for adapting the provenance tracking!

morcuended (approved):
It looks good to me. In the current implementation, you have to make sure to change the DL1b subdirectory prod id in another cfg (e.g. `CatB_tailcut84`) to differentiate it from the original DL1b files (produced only assuming Cat A calibration).

Alternatively, you could implement it in a way that if you activate the application of Cat B, the DL1b subdirectory is automatically changed to include the CatB prefix.

Nonetheless, from my side, this is good to go already.

The failing test is not related, it is just due to the datetime reported in the sequencer table title.

src/osa/scripts/datasequence.py
```diff
@@ -165,6 +174,55 @@ def r0_to_dl1(
...
+        log.debug(f"{run_str} is not the first subrun of the run, so the script "
+            "onsite_create_cat_B_calibration_file will not be launched for this subrun.")
+
+        catB_calibration_file = get_catB_calibration_filename(int(run_str[:5]))
+        n = 0
+        n_max = 10
+        while not catB_calibration_file.exists() and n<=n_max:
+            time.sleep(120)
+            n += 1
+        return 0
```
morcuended, line 203: Here, if the subrun is not the file 0000 the processing will be kept sleeping until the file is available, right? It is not very convenient, but I see it as the most straightforward way in the current OSA scheme in which all analysis steps run within a single batch job for a given subrun.

Setting slurm dependencies would require changing the way OSA submits jobs to something more modular: one job per subrun per analysis step, hence dependencies can be set. In this way, you could also decouple the different time/memory requirements of each step. And maybe everything would be more efficient

morcuended (commented):
I have one concern about adding the creation of CatB and application to DL1a. See comment inline

src/osa/scripts/datasequence.py
morcuended: The way you have implemented it, it can probably work, but I think is prone to problems during processing.
The current scheme has the drawback of not being flexible for mixing processing at subrun and run levels. Normal processing goes subrun-wise, then cat B coefficients calculation is done on a run-wise basis. The same applies to e.g. DL3 production, once you have the DL2 merged run-wise, you would like to run it that way. That's why `datasequence` has this limitation

r0_dl1 (subrun-wise) -> calculate_catB (run-wise)  -> dl1ab (subrun-wise) -> merge dl1 (run-wise) -> dl1_dl2 (run-wise) -> dl2_dl3 (run-wise)

Each step would be a job with dependency on the previous stage.

However, this would be a major change in OSA I think. I do not know where to go from this point

morcuended: probably keep the current implementation and test it for some time

contrera: We discussed it a week ago,, I had the same objections you do have. I suggested implementing dependencies in the individual slurm jobs of an array job, but as Maria pointed out, this  means waiting for the full completion of the 0000 subrun (DL1a, DL1ab). It is clearly a bottle neck.  Is catB calculated only with a subrun or does it use the information all subruns?

morcuended: I think the bottleneck is there anyway with both schemes, you'd have to wait.

Cat B calibration coefficient is calculated runwise, I'd say using the information from all interleaved events across the entire run. If these coefficients are to be applied I see no other way than waiting until the calibration B file is produced. The thing with dependencies is that they automatically link steps, without having to do the sleep until it finishes. Maybe it can be decoupled into two things: first the production of CatB calib files and second the possible application to the data (this is to be checked with Crab data).

I think it is fine as it is. Give it a try.

contrera: For the moment, we can work this way, but later on, I also prefer your proposal. Consider this just another step in the sequencer, not in the data sequence, another calibration step. Build a script that just produces the cat_B calibration for a run. It will depend on the ped_cal calibration, and the data sequence jobs of a given run will depend on the cta_b jobs for this run.

marialainez:
I close this PR since the Cat-B calibration will be implemented as it is done in #302

PR #293, opened 2024-06-12
Improvements for gain selection automation

marialainez:
Main changes that have been made in this PR: 

- Write history files for each subrun (apart from the log files) + update them every time the gain selection script is launched
- Allow launching the gain selection script several times without launching the jobs that are already running again (if the history file exists and is empty for a given subrun, it means that the job is running for that run, so it doesn't launch it again)
- Change the check option so that it is done looking at the history files and not at the log files
- Add simulate option to the gain selection script
- Once the gain selection finishes correctly for all runs, the gain selection script is launched with the check option to see if there is any run missing that needs to be copied to R0G.

### DIFF
--- a/src/osa/configs/sequencer.cfg
+++ b/src/osa/configs/sequencer.cfg
@@ -72,8 +72,8 @@ electron: /path/to/DL2/electron_mc_testing.h5
 PARTITION_PEDCALIB: short, long
 PARTITION_DATA: short, long
 MEMSIZE_PEDCALIB: 3GB
-MEMSIZE_DATA: 16GB
-MEMSIZE_GAINSEL: 15GB
+MEMSIZE_DATA: 6GB
+MEMSIZE_GAINSEL: 2GB
 WALLTIME: 1:15:00
 # Days from current day up to which the jobs are fetched from the queue.
 # Default is None (left empty).
--- a/src/osa/configs/sequencer.cfg
+++ b/src/osa/configs/sequencer.cfg
@@ -55,6 +55,7 @@ dl1_to_dl2: lstchain_dl1_to_dl2
 dl1a_config: /software/lstchain/data/lstchain_standard_config.json
 store_image_dl1ab: True
 merge_dl1_datacheck: True
+use_ff_heuristic_gain_selection: False
 dl1b_config: /software/lstchain/data/lstchain_standard_config.json
 dl2_config: /software/lstchain/data/lstchain_standard_config.json
 rf_models: /data/models/prod5/zenith_20deg/20201023_v0.6.3
--- a/src/osa/scripts/datasequence.py
+++ b/src/osa/scripts/datasequence.py
@@ -6,7 +6,7 @@ from pathlib import Path
 
 from osa.configs import options
 from osa.configs.config import cfg
-from osa.job import historylevel
+from osa.job import historylevel, get_sacct_output, run_sacct
 from osa.workflow.stages import AnalysisStage
 from osa.provenance.capture import trace
 from osa.utils.cliopts import data_sequence_cli_parsing
@@ -285,6 +285,17 @@ def dl1_to_dl2(run_str: str) -> int:
     return analysis_step.rc
 
 
+def is_datasequence_running(run_str: str) -> bool:
+    """Return True if any datasequence jobs are running for the given run number."""
+    sacct_output = run_sacct()
+    sacct_info = get_sacct_output(sacct_output)
+    jobs_run = sacct_info[sacct_info["JobName"]=="LST1_"+run_number]
+    if jobs_run:
+        return True
+    else: 
+        return False
+
+
 def main():
     """Performs the analysis steps to convert raw data into DL2 files."""
     (
@@ -303,18 +314,26 @@ def main():
     else:
         log.setLevel(logging.INFO)
 
-    # Run the routine piping all the analysis steps
-    rc = data_sequence(
-        calibration_file,
-        drs4_ped_file,
-        time_calibration_file,
-        systematic_correction_file,
-        drive_log_file,
-        run_summary_file,
-        pedestal_ids_file,
-        run_number,
-    )
-    sys.exit(rc)
+    if is_datasequence_running(run_number):
+        log.info(
+            f"Jobs launched by datasequence are still running or pending for"
+            "run {run_number}, try again later."
+        )
+        return
+
+    else:
+        # Run the routine piping all the analysis steps
+        rc = data_sequence(
+            calibration_file,
+            drs4_ped_file,
+            time_calibration_file,
+            systematic_correction_file,
+            drive_log_file,
+            run_summary_file,
+            pedestal_ids_file,
+            run_number,
+        )
+        sys.exit(rc)
 
 
 if __name__ == "__main__":
--- a/src/osa/scripts/datasequence.py
+++ b/src/osa/scripts/datasequence.py
@@ -6,7 +6,7 @@ from pathlib import Path
 
 from osa.configs import options
 from osa.configs.config import cfg
-from osa.job import historylevel, get_sacct_output, run_sacct
+from osa.job import historylevel, get_sacct_output
 from osa.workflow.stages import AnalysisStage
 from osa.provenance.capture import trace
 from osa.utils.cliopts import data_sequence_cli_parsing
--- a/src/osa/scripts/datasequence.py
+++ b/src/osa/scripts/datasequence.py
@@ -314,7 +314,7 @@ def main():
     else:
         log.setLevel(logging.INFO)
 
-    if is_datasequence_running(run_number):
+    if not options.simulate and is_datasequence_running(run_number):
         log.info(
             "Jobs launched by datasequence are still running or pending for"
             f"run {run_number}, try again later."
--- a/src/osa/scripts/datasequence.py
+++ b/src/osa/scripts/datasequence.py
@@ -286,11 +286,17 @@ def dl1_to_dl2(run_str: str) -> int:
 
 
 def is_datasequence_running(run_str: str) -> bool:
-    """Return True if any datasequence jobs are running or pending for the given run number."""
+    """Return True if any datasequence jobs are running or pending for the given run number.
+    
+    Parameters
+    ----------
+    run_str : str
+        Run number and subrun number in the format XXXXX.XXXX
+    """
     sacct_output = run_sacct()
     sacct_info = get_sacct_output(sacct_output)
-    jobs_run = sacct_info[sacct_info["JobName"]=="LST1_"+run_str[:5]]
-    if jobs_run:
+    queued_jobs = sacct_info[sacct_info["JobName"]=="LST1_"+run_str[:5]]
+    if queued_jobs:
         return True
     else: 
         return False
--- a/src/osa/scripts/datasequence.py
+++ b/src/osa/scripts/datasequence.py
@@ -6,7 +6,7 @@ from pathlib import Path
 
 from osa.configs import options
 from osa.configs.config import cfg
-from osa.job import historylevel, get_sacct_output
+from osa.job import historylevel
 from osa.workflow.stages import AnalysisStage
 from osa.provenance.capture import trace
 from osa.utils.cliopts import data_sequence_cli_parsing
--- a/src/osa/scripts/datasequence.py
+++ b/src/osa/scripts/datasequence.py
@@ -289,7 +289,7 @@ def is_datasequence_running(run_str: str) -> bool:
     """Return True if any datasequence jobs are running for the given run number."""
     sacct_output = run_sacct()
     sacct_info = get_sacct_output(sacct_output)
-    jobs_run = sacct_info[sacct_info["JobName"]=="LST1_"+run_number]
+    jobs_run = sacct_info[sacct_info["JobName"]=="LST1_"+run_str[:5]]
     if jobs_run:
         return True
     else: 
--- a/src/osa/scripts/datasequence.py
+++ b/src/osa/scripts/datasequence.py
@@ -286,7 +286,7 @@ def dl1_to_dl2(run_str: str) -> int:
 
 
 def is_datasequence_running(run_str: str) -> bool:
-    """Return True if any datasequence jobs are running for the given run number."""
+    """Return True if any datasequence jobs are running or pending for the given run number."""
     sacct_output = run_sacct()
     sacct_info = get_sacct_output(sacct_output)
     jobs_run = sacct_info[sacct_info["JobName"]=="LST1_"+run_str[:5]]
--- a/src/osa/scripts/datasequence.py
+++ b/src/osa/scripts/datasequence.py
@@ -285,23 +285,6 @@ def dl1_to_dl2(run_str: str) -> int:
     return analysis_step.rc
 
 
-def is_datasequence_running(run_str: str) -> bool:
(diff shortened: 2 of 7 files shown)


src/osa/scripts/gain_selection.py
```diff
@@ -134,19 +233,25 @@ def apply_gain_selection(date: str, start: int, end: int, output_basedir: Path =
...
+
     # Apply gain selection only to DATA runs
     data_runs = summary_table[summary_table["run_type"] == "DATA"]
     log.info(f"Found {len(data_runs)} DATA runs to which apply the gain selection")
 
-    output_dir = output_basedir / date
-    log_dir = output_basedir / "log" / date
+    base_dir = Path(cfg.get("LST1", "BASE"))
+    output_dir = base_dir / f"R0G/{date}"
+    log_dir = base_dir / f"R0G/log{date}"
```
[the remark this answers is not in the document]
marialainez: it was not intended, I just changed it :)

src/osa/scripts/gain_selection.py
```diff
@@ -156,98 +261,65 @@ def apply_gain_selection(date: str, start: int, end: int, output_basedir: Path =
...
-    return StringIO(sp.check_output(sacct_cmd).decode())
+
+def job_finished_in_timeout(run, subrun, log_dir):
+
+    job_id = get_last_job_id(run, subrun, log_dir)
+    job_status = get_sacct_output(run_sacct_j(job_id))["State"]
+    if job_status == "TIMEOUT":
+        return True
+    else:
+        return False
```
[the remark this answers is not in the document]
marialainez: I moved the job_finished_in_timeout function to job.py, but the function get_last_job_id searchs the log files of the gain selection, so it's only used when launching the gain_selection script

src/osa/scripts/gain_selection.py
```diff
@@ -126,7 +126,11 @@ def get_sbatch_script(
 
 
 def launch_gainsel_for_data_run(date, run, output_dir, r0_dir, log_dir, log_file, tool, simulate=False):
-    
+    """
+    Create the gain selection sbatch script and launch it for a given run. Runs from before 20231205
+    without UCTS or TIB info are directly copied to the final directory. Subruns that do not have 
+    four streams are also directly copied.
```
morcuended, line 132: Maybe also add parameters in the docstrings and/or at least type hints in each argument (date: str, run: Table, etc)

src/osa/scripts/gain_selection.py
```diff
@@ -88,41 +81,147 @@
...
-        return dedent(
-            f"""\
-        #!/bin/bash
-
-        #SBATCH -D {log_dir}
-        #SBATCH -o "gain_selection_{run_id:05d}_{subrun:04d}_%j.log"
-        #SBATCH --job-name "gain_selection_{run_id:05d}"
-        #SBATCH --mem=40GB
-        #SBATCH --partition=short,long
+        mem_per_job = cfg.get("SLURM", "MEMSIZE_GAINSEL")
```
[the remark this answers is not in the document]
morcuended: Yes, so you're able to control it

src/osa/scripts/gain_selection.py
```diff
@@ -263,64 +371,90 @@ def GainSel_finished(date: str) -> bool:
...
+            if gainsel_rc == "1":
+                log.warning(f"Gain selection failed for run {run_id}.{subrun}")
+                failed_subruns.append(file)
+
+            elif gainsel_rc == "0":
+                log.debug(f"Gain selection finished successfully for run {run_id}.{subrun}")
+        else:
+            log.info(f"Gain selection is still running for run {run_id}.{subrun}")
+ 
+    if failed_subruns:
```
morcuended, line 397: Maybe worth logging these failed_subruns as well?

src/osa/scripts/gain_selection.py
```diff
@@ -134,19 +260,26 @@ def apply_gain_selection(date: str, start: int, end: int, output_basedir: Path =
...
-    output_dir.mkdir(parents=True, exist_ok=True)
-    log_dir.mkdir(parents=True, exist_ok=True)
+    base_dir = Path(cfg.get("LST1", "BASE"))
+    r0_dir = base_dir / "R0" / date
+    output_dir = base_dir / f"R0G/{date}"
+    log_dir = base_dir / f"R0G/log/{date}"
     log_file = log_dir / f"r0_to_r0g_{date}.log"
-    r0_dir = Path(f"/fefs/aswg/data/real/R0/{date}")
+    if not simulate:
+        output_dir.mkdir(parents=True, exist_ok=True)
```
morcuended, line 281: although it is more clear with this line, it is not needed. The line below `log_dir` will create the parent output_dir directory

morcuended (commented):
Here are some more comments. Looks good overall. IMO some parts are too nested, but I understand that there are different cases you have to cope with

src/osa/scripts/gain_selection.py
```diff
@@ -263,64 +371,90 @@ def GainSel_finished(date: str) -> bool:
...
-    log_dir = output_basedir / "log" / date
-    filenames = glob.glob(f"{log_dir}/gain_selection*.log")
-    jobs = [re.search(r'(?<=_)(.[0-9.]+?)(?=.log)', i).group(0) for i in filenames]
+    base_dir = Path(cfg.get("LST1", "BASE"))
+    log_dir = base_dir / f"R0G/log/{date}"
+    history_files = log_dir.glob(f"gain_selection_{run_id:05d}.????.history")
+    failed_subruns = []
+    log.info(f"Checking all history files of run {run_id}")
+    
+    for file in list(history_files):
```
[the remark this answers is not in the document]
marialainez: not really, I removed it

src/osa/scripts/gain_selection.py
marialainez: what do you mean? there are logging messages after this line

src/osa/scripts/gain_selection.py
```diff
@@ -263,64 +371,90 @@ def GainSel_finished(date: str) -> bool:
...
-            failed_jobs.append(job)
+def check_failed_jobs(date: str):
+    """Search for failed jobs in the log directory."""
 
-    if failed_jobs:
-        log.warning(f"{date}: some jobs did not finish successfully")
+    run_summary_dir = Path(cfg.get("LST1", "RUN_SUMMARY_DIR"))
+    run_summary_file = run_summary_dir / f"RunSummary_{date}.ecsv"
+    summary_table = Table.read(run_summary_file)
+    data_runs = summary_table[summary_table["run_type"] == "DATA"]
```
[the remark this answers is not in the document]
marialainez: the summary_table is also need it in those cases, so I created a function for it

src/osa/scripts/gain_selection.py
morcuended: yes, but you only mention some failed jobs for a given run. Isn't it useful to know which subruns as well?

marialainez: yes, but the subrun number of the failed job is also said above (in line 394)

morcuended: ok!

src/osa/scripts/gain_selection.py
```diff
@@ -311,6 +309,14 @@ def apply_gain_selection(date: str, start: int, end: int, tool: str = None, no_q
                     sp.run(["cp", file, output_dir])
 
 
+def run_summary_table(date: str) -> Table:
```
morcuended, line 311: there should be a similar function defined somewhere in the code already that might be used for this:

https://github.com/cta-observatory/lstosa/blob/e148c8c61fcec0f51cbdf1399f042a4fb0b302bd/src/osa/nightsummary/nightsummary.py#L47-L79

morcuended:
@marialainez, this would allow running the gain selection launcher several times and just re-doing jobs that did not previously succeed, right?

marialainez:
> @marialainez, this would allow running the gain selection launcher several times and just re-doing jobs that did not previously succeed, right?

yes, we have been testing it for the last days and it works as expected

src/osa/scripts/gain_selection.py
contrera: One more comment: I was thinking about building a web page to watch the gain_selection operation, as we do for the sequencer. This failed_runs list could be used to compute the percentage of advance and build it.

morcuended: Good idea. Maybe the check option should export a table object that can be easily formatted and uploaded to the web with this information

src/osa/scripts/gain_selection.py
marialainez: I have just noticed now that the line 281 (output_dir.mkdir(parents=True, exist_ok=True)) is necessary, since output_dir is not a parent directory of log_dir.

src/osa/scripts/datasequence.py
```diff
@@ -285,6 +285,17 @@ def dl1_to_dl2(run_str: str) -> int:
     return analysis_step.rc
 
 
+def is_datasequence_running(run_str: str) -> bool:
+    """Return True if any datasequence jobs are running for the given run number."""
+    sacct_output = run_sacct()
+    sacct_info = get_sacct_output(sacct_output)
+    jobs_run = sacct_info[sacct_info["JobName"]=="LST1_"+run_str[:5]]
```
morcuended, line 292: running or pending, it should be for both (any job in the queue for that run number), right?

src/osa/scripts/gain_selection.py
```diff
@@ -156,171 +292,176 @@ def apply_gain_selection(date: str, start: int, end: int, output_basedir: Path =
...
+
+            if gainsel_rc == "1":
+                log.warning(f"Gain selection failed for run {run_id}.{subrun}")
+                failed_subruns.append(file)
+
+            elif gainsel_rc == "0":
+                log.debug(f"Gain selection finished successfully for run {run_id}.{subrun}")
+        else:
+            log.info(f"Gain selection is still running for run {run_id}.{subrun}")
+            return False
```
morcuended, line 399: this is somewhat similar to the function in webmaker for gain selection, but can be simplified later

src/osa/scripts/gain_selection.py
```diff
@@ -156,171 +292,176 @@ def apply_gain_selection(date: str, start: int, end: int, output_basedir: Path =
...
-            )
+    for run in data_runs:
+        run_id = run["run_id"]
+        
+        if not check_gainsel_jobs_runwise(date, run_id):
+            log.warning(f"Gain selection did not finish successfully for run {run_id}.")
+            failed_runs.append(run)
+
+    if failed_runs:
+        log.info(f"Gain selection did not finish successfully for {date_to_iso(date)}, cannot create the flag file.")
```
morcuended, line 426: maybe warning?

morcuended (commented):
looks good overall, if you have tested it in the cluster for a couple of days I'd say it's good to go

src/osa/scripts/datasequence.py
marialainez: yes

morcuended: then modify the docstring accordingly, maybe refer to submitted (running/pending) jobs instead of just running or jobs in the queue?

src/osa/scripts/datasequence.py
```diff
@@ -286,7 +286,7 @@ def dl1_to_dl2(run_str: str) -> int:
 
 
 def is_datasequence_running(run_str: str) -> bool:
-    """Return True if any datasequence jobs are running for the given run number."""
+    """Return True if any datasequence jobs are running or pending for the given run number."""
     sacct_output = run_sacct()
     sacct_info = get_sacct_output(sacct_output)
     jobs_run = sacct_info[sacct_info["JobName"]=="LST1_"+run_str[:5]]
```
morcuended, line 292:
```suggestion
    queued_jobs = sacct_info[sacct_info["JobName"]=="LST1_"+run_str[:5]]
```

src/osa/scripts/datasequence.py
```diff
@@ -285,6 +285,17 @@ def dl1_to_dl2(run_str: str) -> int:
     return analysis_step.rc
 
 
+def is_datasequence_running(run_str: str) -> bool:
+    """Return True if any datasequence jobs are running or pending for the given run number."""
```
morcuended, line 289:
```suggestion
    """Return True if any datasequence jobs are running or pending for the given run number.
    
    Parameters
    ----------
    run_str : str
        Run number and subrun number in the format XXXXX.XXXX
    """
```

src/osa/scripts/gain_selection.py
```diff
@@ -422,7 +437,10 @@ def check_failed_jobs(date: datetime):
             log.warning(f"Gain selection did not finish successfully for run {run_id}.")
             failed_runs.append(run)
 
-    if failed_runs:
+        if not check_warnings_in_logs(date, run_id):
+            warnings.append(run)
```
morcuended, line 442: better use `warning_in_logs` as name of this variable

src/osa/scripts/gain_selection.py
```diff
@@ -408,12 +408,27 @@ def check_gainsel_jobs_runwise(date: datetime, run_id: int) -> bool:
         return True
 
 
+def check_warnings_in_logs(date: datetime, run_id: int) -> bool:
+    """Look for warnings in the log files created by lstchain_r0_to_r0g."""
+    log_dir = base_dir / f"R0G/log/{date_to_dir(date)}"
+    log_files = log_dir.glob(f"r0_to_r0g_{run_id:05d}.*.log")
+    for file in log_files:
+        content = file.read_text().splitlines()
+        for line in content:
```
morcuended, line 418: are the log files usually large? this check could take time

morcuended (commented):
some changes in the log messages, to make easier to follow the logs

src/osa/scripts/gain_selection.py
```diff
@@ -417,8 +417,8 @@ def check_warnings_in_logs(date: datetime, run_id: int) -> bool:
     for file in log_files:
         content = file.read_text().splitlines()
         for line in content:
-            if "WARNING" in line:
-                log.warning(f"There is a warning in the log files of run {run_id}: {line}")
+            if "WARNING - You should use heuristic identification of FF events!" in line:
+                log.warning(f"You should use heuristic identification of FF events for run {run_id}!")
```
morcuended, line 419: I wonder in this case, would it be easy to pass this information to the gain selection command builder to add the ff heuristic and run it or prepare it for a second execution with that option?

morcuended: I guess then this would be problematic in subsequent R0 to DL1 processing, since I think you have to be coherent in the usage of FF heuristic option

src/osa/configs/sequencer.cfg
```diff
@@ -71,6 +72,7 @@ PARTITION_PEDCALIB: short, long
 PARTITION_DATA: short, long
 MEMSIZE_PEDCALIB: 3GB
 MEMSIZE_DATA: 16GB
+MEMSIZE_GAINSEL: 15GB
```
morcuended, line 75: even if this is a dummy cfg, in the actual cfg in use, this mem req can be reduced, now that the memory leak was fixed in protozfit

morcuended (commented):
Some more comments (I hope the last ones)

morcuended:
A problem with the mamba setup step in CI makes tests fail. Please relaunch the tests after #303.

morcuended:
@marialainez please rebase the branch with main to make the tests run again

PR #297, opened 2024-07-10
Update gainsel_webmaker.py

rcervinoucm:
Small changes in gain selection webmaker


src/osa/scripts/gainsel_webmaker.py
```diff
@@ -94,6 +94,8 @@ def determine_status(row):
         if row["failed"] > 0:
             return "FAILED"
         elif row["pending"] > 0:
+            return "RUNNING"
```
morcuended, line 97: I think it'll never reach the pending with these clauses.

This has to be the second clause
```python
elif row["pending"] == row["n_subruns"]:
    return "PENDING"
```
3rd the COMPLETED case, which are more restrictive.
Then the running one, finally not started

morcuended (approved):

PR #302, opened 2024-10-14
Application of Cat-B calibration (option1)

marialainez:
In order to be able to process some nights "fast" with Cat-B calibration we have thought of this option. The idea is to launch first a sequencer with option --no-dl1ab, then the script onsite_create_cat_B_calibration "manually" and then another sequencer (without option --no-dl1ab). I have created a script that does all these steps by its own and it seems to be working. Tests are ongoing.


src/osa/scripts/datasequence.py
```diff
@@ -68,13 +72,17 @@ def data_sequence(
...
+            level = 0
+            log.info(f"No DL1B are going to be produced. Going to level {level}")
+        else: 
+            rc = dl1ab(run_str)
+            if cfg.getboolean("lstchain", "store_image_dl1ab"):
+                level -= 1
+                log.info(f"Going to level {level}")
+            else:
+                level -= 2
+                log.info(f"No images stored in dl1ab. Producing DL2. Going to level {level}")
```
morcuended, line 85:
```suggestion
                log.info(
                    f"No images stored in dl1ab. Skipping datachecks and producing DL2. Going to level {level}"
                )
```

src/osa/scripts/datasequence.py
```diff
@@ -165,6 +173,55 @@ def r0_to_dl1(
...
+        log.debug(f"{run_str} is not the first subrun of the run, so the script "
+            "onsite_create_cat_B_calibration_file will not be launched for this subrun.")
+
+        catB_calibration_file = get_catB_calibration_filename(int(run_str[:5]))
+        n = 0
+        n_max = 10
+        while not catB_calibration_file.exists() and n<=n_max:
+            time.sleep(120)
+            n += 1
+        return 0
```
morcuended, line 202: Do you need the sleep part here if you plan to do it in two steps?

morcuended, line 202: I think I'd simply return if subrun is not 0

src/osa/job.py
```diff
@@ -215,6 +221,8 @@ def historylevel(history_file: Path, data_type: str):
                 # Data sequence
                 elif program == cfg.get("lstchain", "r0_to_dl1"):
                     level = 3 if exit_status == 0 else 4
+                #elif program == cfg.get("lstchain", "catB_calibration"):
+                #    level = 3 if exit_status == 0 else 4
```
morcuended, line 225: why is it commented? do you plan to include a new cat B calibration line in the history file?

marialainez:
The idea with this new scheme is that first one sequencer is launched with option --no-dl1ab. This sequencer will be launched once by the crontab. Then sequencer_catB_tailcuts will be launched several times. It has to be launched every x minutes to give time for the catB.closed files and the .json config files with the appropriate tailcuts to be created. Then, a second sequencer (without the option --no-dl1ab) is launched also every x minutes. This sequencer should start from step dl1ab and it should first check if the catB.closed files and the .json config files for the tailcuts exist for a given run. Once it has checked that these files exist, it will proceed with the dl1ab step and dl2 if option --no-dl2 is not given.

marialainez:
> Will this be the final implementation instead of #292? Should we close #292 then?

Yes, we can close #292

PR #310, opened 2024-10-25
Use lstcam_calib instead of lstchain for the calibration

marialainez:


morcuended:
if command names are defined via cfg, the change is quite straightforward, just adapting the test to expect whatever command name is included in the test cfg

morcuended (changes requested):
Thanks @marialainez.

Try to use cfg.get in tests to get the command name expected. Otherwise looks good, the change is quite straightforward

environment.yml
```diff
@@ -27,6 +27,7 @@ dependencies:
   - pyirf~=0.10
   - lstchain>=0.10.7
   - tenacity
+  - lstcam_calib
```
morcuended, line 30:
```suggestion
  - pip:
      - lstcam_calib
```

morcuended:
ctapipe_io_lst v0.24 ready in conda forge already, running tests again with lstcam calib here

morcuended:
I realized that to make the CI work with lstcam_calib and lstchain, lstosa should require lstchain >= 0.10.13 (which is the release that requires ctapipe_io_lst >= 0.24.0)

environment.yml
```diff
@@ -25,8 +25,10 @@ dependencies:
   - ctapipe_io_lst=0.22
```
morcuended, line 25:
```suggestion
  - ctapipe_io_lst=0.24
```

morcuended:
@marialainez, @rcervinoucm please modify tests accordingly

morcuended (approved):
Looks good

PR #313, opened 2024-11-04
Check if history files of a run exist before creating .closed files

marialainez:


src/osa/scripts/gain_selection.py
```diff
@@ -380,13 +380,20 @@ def GainSel_finished(date: datetime) -> bool:
...
     return flagfile.exists()
-
+   
 
 def check_gainsel_jobs_runwise(date: datetime, run_id: int) -> bool:
     """Search for failed jobs in the log directory."""
     base_dir = Path(cfg.get("LST1", "BASE"))
     log_dir = base_dir / f"R0G/log/{date_to_dir(date)}"
     history_files = log_dir.glob(f"gain_selection_{run_id:05d}.????.history")
+    summary_table = run_summary_table(date)
```
morcuended, line 390: history_files gives you a iterator object that can be used once. Make sure that history_files iterator can be reused later when you do `for file in history_files`. Otherwise you might need to define it in the first place as a list:

```python
history_files = list(log_dir.glob(f"gain_selection_{run_id:05d}.????.history"))
```

morcuended (approved):
looks good now, thanks!

Issue #229, opened 2023-06-10
Prod_id wrongly reported by check point method


morcuended:
https://github.com/cta-observatory/lstosa/blob/799fc14669b07d182bdfd8a24dce916e0034e714/osa/workflow/stages.py#L106-L115

`prod_id` should indicate the corresponding prod ID of each data level (e.g., `dl1_prod_id`), not always the general `options.prod_id`

PR #315, opened 2024-11-05
Write the prod_id corresponding to each step in the history files

marialainez:
Solves issue #229

### DIFF
--- a/src/osa/workflow/stages.py
+++ b/src/osa/workflow/stages.py
@@ -121,9 +121,14 @@ class AnalysisStage:
 
     def _write_checkpoint(self):
         """Write the checkpoint in the history file."""
+        if self.command in ["lstchain_data_r0_to_dl1", "lstchain_dl1ab", "lstchain_check_dl1"]: 
+            prod_id = options.dl1_prod_id
+        elif self.command == "lstchain_dl1_to_dl2":
+            prod_id = options.dl2_prod_id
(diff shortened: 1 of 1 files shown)


src/osa/workflow/stages.py
```diff
@@ -121,9 +122,19 @@ def _remove_dl1b_output(self, file_prefix):
 
     def _write_checkpoint(self):
         """Write the checkpoint in the history file."""
+        dl1_commands = [
+            cfg.get("lstchain", "r0_to_dl1"),
+            cfg.get("lstchain", "dl1ab"),
+            cfg.get("lstchain", "check_dl1")
+            ]
```
morcuended, line 128: might be better to create a dictionary matching the command to the prod id:

```python
command_to_prod_id = {
    cfg.get("lstchain", "r0_to_dl1"): options.prod_id,
    cfg.get("lstchain", "dl1ab"): options.dl1_prod_id,
    cfg.get("lstchain", "check_dl1"): options.dl1_prod_id,
    cfg.get("lstchain", "dl1_to_dl2"): options.dl2_prod_id
}
```
then:

```python
prod_id = command_to_prod_id.get(self.command)
```

morcuended (changes requested):

morcuended (approved):
Looks good, thanks!

PR #317, opened 2024-11-08
Use the SLURM account specified in the cfg also for the gainsel and autocloser jobs 

marialainez:


contrera (approved):

morcuended (approved):
good that spotted this!

PR #327, opened 2025-06-17
Update to lstchain v0.11

marialainez:


marialainez:
Tests are failing due to incompatibilities between lstchain and lstcam_calib. lstcam_calib requires ctapipe >=0.19.3,<0.20.0 
while lstchain >=0.11.0 requires ctapipe >=0.25.0. 

@FrancaCassol can ctapipe be upgraded in lstcam_calib?

moralejo:
If that is not possible, for whatever reason, what about this suggestion from ChatGPT? 

> You can call a script in another environment from your main environment using subprocess:


> ```
> import subprocess
> subprocess.run(["conda", "run", "-n", "env2", "python", "script_using_packageB.py"])
> ```

You could have lstcam_calib under its own environment. This would make it unnecessary to keep the dependencies consistent.

marialainez:
> Yes, please, use a different environment for lstcam_calib. We do not plan to follow the lschain environment changes.

Ok, we will do it like that

marialainez:
The test is failing because lstcam_calib is not installed in the environment. I merge the PR.

PR #330, opened 2025-06-18
Consider only files produced with version vX.Y.Z to create the symlink

marialainez:


marialainez:
The failure in the tests is unrelated, and it's solved now in #327, so I merge this PR.

PR #333, opened 2025-06-23
Change calibration directory names following names in the cfg

marialainez:


marialainez:
Unrelated error, I merge PR.

Issue #346, opened 2025-07-24
Sequencer is launched again if the dl1_to_dl2 jobs launched by autocloser fail


marialainez:
If any job launched by autocloser to perform the dl1_to_dl2 step fails, then the function "are_all_jobs_correctly_finished" returns False, and "is_sequencer_completed" returns False. Therefore, sequencer launches all the sequences again (even if the DL1ab files were already correctly produced for all runs). I think a new way of defining "is_sequencer_completed" should be found, so that it doesn't take into account the DL2 files production.

marialainez:
This does not only happen if any dl1_to_dl2 job fails, but always. The function "are_all_jobs_correctly_finished" returns always False because we are currently not writing the dl2 step in the .history files, so unless the --no-dl2 option is used, sequencer thinks all subruns are always incompleted. Probably the easiest solution currently is to change "if out == 0:" to "if out == 1:" in line 118 of job.py, because no dl2 step is written in the .history files.

PR #350, opened 2025-07-30
Check if sequences are completed only up to level DL1AB

marialainez:
see issue #346

### DIFF
--- a/src/osa/job.py
+++ b/src/osa/job.py
@@ -115,7 +115,7 @@ def are_all_jobs_correctly_finished(sequence_list):
             # we need to check all the subrun wise history files
             # .../sequence_LST1_04180.XXXX.history
             out, _ = historylevel(history_file, sequence.type)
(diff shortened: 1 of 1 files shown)


marialainez:
I was completely wrong about this. I thought that "historylevel" was returning 0 only when the dl2 step was completed, but that was already changed when the dl2 production was moved to autocloser. So currently historylevel returns 0 if the dl1ab step and the datacheck production finished successfully, which is what we are expecting. What was really happening is that for some reason (we have to check why) no history files are being produced for the calibration sequences, and therefore, the function "are_all_jobs_correctly_finished" returns False (see line 109 of job.py). I have adapted this so that only the data sequences are considered for checking if sequencer processing is already completed. I have made a small test and it works as expected.

marialainez:
We have tested this with tonight's data and it worked as expected, so I merge the PR.

PR #359, opened 2025-09-26
Add Cat-B status to sequencer status table

rcervinoucm:


src/osa/scripts/sequencer.py
```diff
@@ -205,6 +206,36 @@ def update_sequence_status(seq_list):
...
+            seq.catbstatus = check_catB_status(seq)
+
+
+def check_catB_status(seq):
+    catbstatus = 'None'
+    if seq.type == "DATA":
+        directory = options.directory
+        # Buscar archivos "closed"
+        closed_files = list(directory.glob(f"catB*{seq.run}*.closed"))
+        if any(f.exists() for f in closed_files):
```
marialainez, line 218: This shouldn't be necessary since only one .closed file is created per run, so I think simply "if closed_file.exists()" should work.

src/osa/scripts/sequencer.py
```diff
@@ -205,6 +206,36 @@ def update_sequence_status(seq_list):
...
+
+            if log_files and any(f.exists() for f in log_files):
+                status_set = False
+                for f in log_files:
+                    if f.exists():
+                        with f.open("r") as fh:
+                            for line in fh:
+                                if "CRITICAL" in line:
+                                    seq.catbstatus = 'FAILED'
+                                    status_set = True
```
marialainez, line 231: I think it would be easier to get status of the SLURM job with sacct, in lstosa we usually use these functions: get_sacct_output(run_sacct(job_id=job_id))

src/osa/scripts/sequencer.py
```diff
@@ -205,6 +206,36 @@ def update_sequence_status(seq_list):
...
+
+
+                # Si ningún log tenía CRITICAL/ERROR pero existen archivos
+                if not status_set and any(f.exists() for f in log_files):
+                    catbstatus = 'RUNNING'
```
marialainez, line 236: The same here, if you get the SLURM job id, then using get_sacct_output(run_sacct(job_id=job_id)) you can get its status.

marialainez (changes requested):

src/osa/scripts/sequencer.py
rcervinoucm: You are right!

src/osa/scripts/sequencer.py
rcervinoucm: Changed! Thanks for the review

marialainez (approved):

PR #377, opened 2026-03-09
Add NightFinished flag to launchers

mariahri-ucm:
Update of the sequencer and autocloser launchers to add the NightFinished flag, preventing them from running when the night is already closed.


contrera (approved):
Checked and agreed

Issue #389, opened 2026-05-23
Error Additional NSB search


alexcervino:
We found an issue on sequencer 2 regarding that the script looks for the Additional NSB rate in the TailCuts logs for each run, but is not ready to look for values under 0.

PR #390, opened 2026-05-23
Fix regex to capture negative NSB rates

alexcervino:
Regular expression fix related to issue #389

### DIFF
--- a/src/osa/utils/utils.py
+++ b/src/osa/utils/utils.py
@@ -332,7 +332,7 @@ def get_nsb_level(run_id):
     log_file = tailcuts_finder_dir / f"log_find_tailcuts_Run{run_id:05d}.log"
     with open(log_file, "r") as file:
         log_content = file.read()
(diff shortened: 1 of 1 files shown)


src/osa/utils/utils.py
```diff
@@ -414,4 +414,4 @@ def get_lstcam_calib_version(env_path: Path) -> str:
     result = sp.run(cmd, capture_output=True, text=True, check=True)
     for line in result.stdout.split('\n'):
         if line.startswith('Version:'):
-            return line.split(':', 1)[1].strip()
```
contrera, line 417: I don´t understand this change

contrera (commented):
Good, ¡thanks!

Perhaps we should raise an error when this happens, as it shouldn´t happen normally.

alexcervino:
Thanks a lot for the review @contrera , I have set an error when NSB is below 0.

contrera (approved):
Good, lets go on

PR #397, opened 2026-07-22
 CatB/Tailcuts pilot script

Manuel19032002:
This PR introduces a sequencer → pilot → worker workflow for CatB/Tailcuts processing. Instead of running CatB and Tailcuts directly from the sequencer, sequencer_catB_tailcuts.py now submits pilot jobs that execute catb_tailcuts_pipeline.py on a run-by-run basis.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,6 +77,7 @@ gain_selection = "osa.scripts.gain_selection:main"
 update_source_catalog = "osa.scripts.update_source_catalog:main"
 gainsel_webmaker = "osa.scripts.gainsel_webmaker:main"
 sequencer_catB_tailcuts = "osa.scripts.sequencer_catB_tailcuts:main"
+catb_tailcuts_pipeline = "osa.scripts.catb_tailcuts_pipeline:main"
 
 [tool.setuptools.packages.find]
 where = ["src"]
--- a/src/osa/catb_utils.py
+++ b/src/osa/catb_utils.py
@@ -1,0 +1,115 @@
+import logging
+import re
+from pathlib import Path
+
+from osa.configs import options
+from osa.configs.config import cfg
+from osa.utils.logging import myLogger
+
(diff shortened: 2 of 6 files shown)


contrera (commented):
This is what copilot thinks

Thanks for the refactor — moving CatB/Tailcuts execution out of the sequencer into pilot-submitted run-level workers is a strong architectural improvement.

✅ What looks good
Clearer separation of responsibilities (sequencer orchestration vs worker execution).
Better scaling potential with run-by-run parallelism.
Better fault isolation compared to direct execution from the sequencer.
🔎 Approval conditions (before merge)
I’m supportive, but I’d like the following to be explicitly covered/confirmed:

Idempotency / duplicate submission safety
Re-running or re-submitting the same run should not create duplicate or inconsistent outputs.

Failure propagation and terminal states
Worker failures must deterministically propagate back to pilot/sequencer, with clear final states (success/failed) per run.

Retry behavior is explicit and bounded
Retry policy should be documented in code/config (max attempts, backoff, terminal failure behavior), avoiding unbounded retries.

Timeout/cancellation handling
Pilot/worker jobs should enforce execution timeouts and handle cancellation cleanly (no orphaned state).

Observability and traceability
Logs should consistently include correlation keys (at least run ID + job ID) across sequencer/pilot/worker so incidents can be traced end-to-end.

CI/test evidence for non-happy paths
Please include test evidence (or a follow-up issue) covering at least:

success path
transient failure with retry and eventual success
permanent failure with correct terminal handling
💡 Suggested follow-ups (can be separate PRs if needed)
Add a short architecture note describing the sequencer/pilot/worker contract and state transitions.
Add/expand integration tests for retry/failure/cancellation semantics.
Add lightweight metrics (queue depth, run latency, retry count, failure rate) for operational monitoring.
If the above points are addressed (or clearly tracked with near-term follow-ups), I’d be comfortable approving.

PR #400, opened 2026-07-29
Add version specification for Ruff action

Manuel19032002:


contrera (approved):
Last version of Ruff seems too strict, this is going back to an earlier one

